PR #1, opened 2023-11-28
Add pre-commit to repo.

manugoyal:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+    rev: v2.2.5
+    hooks:
+      - id: codespell
+        exclude: >
+          (?x)^(
+              .*\.(json|prisma|svg)
+          )$
+        args: ["-L rouge,coo,couldn,unsecure"]
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.7.1
```
ankrgyl, line 17: since this is new, could we use the newer version of prettier (3.1.0)?

apis/vercel/pnpm-lock.yaml
ankrgyl, line 1: We should exclude this file from pre-commit hooks (it gets expensive to reformat it on every commit)

pnpm-lock.yaml
ankrgyl, line 1: we should exclude this file

apis/vercel/pnpm-lock.yaml
ankrgyl: also i _think_ we might be able to delete this file

manugoyal: I removed it from the hooks for now. Will try deleting it in another PR.

manugoyal: nvm just did it in this PR. Seems to work locally

PR #3, opened 2023-12-06
Enable customized caching in local proxy

manugoyal:
For unit testing, it is convenient to be able to control the lifetime and contents of the cache precisely. To accomplish this, we allow users to specify a `BRAINTRUST_PROXY_LOCAL_CACHE_PATH` environment variable, pointing to a JSON file of the cache contents. If specified, the local nodeJS version of the proxy will use this as the cache instead of redis, and supports dumping the cache contents back to this file at any point in time.

Other changes are just cleaning up some function signatures and passing more hooks through to control caching behavior.

### DIFF
--- a/apis/node/src/local.ts
+++ b/apis/node/src/local.ts
@@ -1,3 +1,4 @@
+import fs from "fs";
 import express, { Response } from "express";
 import dotenv from "dotenv";
 import cors from "cors";
@@ -5,7 +6,9 @@ import { pipeline } from "stream/promises";
 
 import { completion } from "./ai";
 import { nodeProxyV1 } from "./node-proxy";
-import { resetEnv } from "./env";
+import { Env, resetEnv } from "./env";
+
+import { CacheKeyOptions } from "@braintrust/proxy";
 
 dotenv.config({ path: ".env.local" });
 resetEnv();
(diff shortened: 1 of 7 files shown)


apis/node/src/local.ts
```diff
@@ -22,7 +25,51 @@ function processError(res: Response, err: any) {
...
+let _testingCache: Record<string, string> | null = null;
+async function loadTestingCache() {
+  if (_testingCache === null) {
+    const contents = await fs.promises.readFile(Env.localCachePath, {
+      encoding: "utf-8",
+    });
+    _testingCache = JSON.parse(contents) as Record<string, string>;
+  }
+  return _testingCache;
+}
```
ankrgyl, line 41: I think this would probably be much cleaner implemented in the main braintrust repo as a separate server (using node) that happens to use these cache functions, vs. littering the main implementation.

Of note, any additional logic we place here bloats the bundle that gets deployed to lambda

ankrgyl (commented):
LGTM but think we should prob split or refactor so that the "testing cache" is a diff implementation vs. conditional logic that pollutes this node server

manugoyal:
> LGTM but think we should prob split or refactor so that the "testing cache" is a diff implementation vs. conditional logic that pollutes this node server

Makes sense. I will work on moving the "testing" version of the server to a separate implementation in the internal repo.

manugoyal:
Ok I removed the testing-related hooks into this package. Now it's mostly just cleanups, and the only real addition is a `cacheKeyOptions` parameter which controls how the cache key digest is created.

packages/proxy/src/proxy.ts
```diff
@@ -41,27 +41,48 @@ const CREDS_CACHE_HEADER = "x-bt-use-creds-cache";
...
+  body,
+  setHeader,
+  setStatusCode,
+  res,
+  getApiSecrets,
+  cacheGet,
+  cachePut,
+  digest,
+  meterProvider = NOOP_METER_PROVIDER,
+  cacheKeyOptions,
```
ankrgyl, line 65: maybe default it to `{}`?

packages/proxy/src/proxy.ts
```diff
@@ -74,20 +95,17 @@ export async function proxyV1(
...
+    Object.entries(proxyHeaders).filter(
+      ([h, _]) =>
+        !(
+          h.startsWith("x-amzn") ||
+          h == CACHE_HEADER ||
+          h == CREDS_CACHE_HEADER ||
+          h == "content-length"
+        ),
+    ),
+  );
```
ankrgyl, line 108: Is there any change here?

ankrgyl (approved):

packages/proxy/src/proxy.ts
manugoyal: not logically. Just code cleanup.

PR #12, opened 2023-12-17
Implement tracing inside the proxy.

manugoyal:
Linked to https://github.com/braintrustdata/braintrust-sdk/pull/65.


packages/proxy/package.json
```diff
@@ -74,6 +74,7 @@
     "@opentelemetry/resources": "^1.18.1",
     "@opentelemetry/sdk-metrics": "^1.18.1",
     "ai": "2.2.22",
+    "braintrust": "link:../../../sdk/js",
```
manugoyal, line 77: Will need to set this to the correct SDK version once we land https://github.com/braintrustdata/braintrust-sdk/pull/65.

PR #17, opened 2024-01-15
Add BRAINTRUST_API_URL to wrangler template.

manugoyal:


manugoyal:
Instead of doing it this way (which will mess up proxy users who try to copy `wrangler-template.toml` to `wrangler.toml`), we will use our own `wrangler.toml` file inside our repo for local services.

PR #19, opened 2024-01-27
Add support for custom models

ankrgyl:
This change accompanies some mainline Braintrust changes which allow you to query secrets by model name (e.g. `gpt-3.5-myco`), rather than endpoint type (e.g. `openai`), and then use additional `customModel` metadata to query the models.

Users can leverage this functionality by configuring models with custom secret types, and then passing the configured model names down to the proxy.


packages/proxy/schema/secrets.ts
```diff
@@ -0,0 +1,74 @@
+import { z } from "zod";
+
+export const BaseMetadataSchema = z
```
ankrgyl, line 3: This wasn't strictly necessary, but the validation logic for model metadata is getting quite interesting, and I figured it was a good time to upgrade it to Zod.

manugoyal, line 3: This one doesn't need to be exported? Also what do you think the naming convention for schema objects should be? I have been declaring them like normal variables, so this would be `baseMetadataSchema`, but I can see the argument for naming them like types because they're basically encoding type information.

packages/proxy/schema/secrets.ts
```diff
@@ -0,0 +1,74 @@
...
+    name: z.string().optional(),
+    secret: z.string(),
+    metadata: z.record(z.unknown()).optional(),
+  })
+  .strict();
+
+export const APISecretSchema = z.union([
+  APISecretBaseSchema.merge(
+    z.object({
+      type: z.union([
```
manugoyal, line 48: isn't there `z.enum` for this?

manugoyal (approved):

packages/proxy/schema/secrets.ts
ankrgyl: fixed

packages/proxy/schema/secrets.ts
ankrgyl: Yeah I personally like capitalizing global constants and types, and this is certainly the former and arguably the latter, so prefer capitalizing them.

PR #31, opened 2024-03-12
Add support for multimodal message content

aphinx:
Adds support for multimodal content, currently just for `gpt-4-vision-preview` and the `claude-3` variants.

### DIFF
--- a/packages/proxy/schema/index.ts
+++ b/packages/proxy/schema/index.ts
@@ -1,3 +1,4 @@
+import { z } from "zod";
 import type {
   AnyModelParam,
   Message,
@@ -28,6 +29,7 @@ export type ModelEndpointType = (typeof ModelEndpointType)[number];
 export interface ModelSpec {
   format: ModelFormat;
   flavor: PromptInputType;
+  multimodal?: boolean;
 }
 
 export const MessageTypes: { [name in ModelFormat]: MessageRole[] } = {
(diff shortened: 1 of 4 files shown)


packages/proxy/schema/index.ts
```diff
@@ -28,6 +29,7 @@ export type ModelEndpointType = (typeof ModelEndpointType)[number];
 export interface ModelSpec {
   format: ModelFormat;
   flavor: PromptInputType;
+  multimodal?: boolean;
```
ankrgyl, line 32: We should update `BaseMetadataSchema` in proxy/schema/secret.ts as well, so that users can input multimodal models into their list

ankrgyl (approved):

packages/proxy/schema/secrets.ts
```diff
@@ -8,7 +8,7 @@ export const BaseMetadataSchema = z
         z.object({
           format: z.enum(["openai", "anthropic", "google"]),
           flavor: z.enum(["completion", "chat"]),
-          multimodal: z.boolean().nullish(),
+          multimodal: z.boolean().optional(),
```
ankrgyl, line 11: Why not use `nullish()`? It's generally safer (especially when writing data to a database)

benigeri: Let's ship!

PR #37, opened 2024-03-28
Send exact origin (matching whitelist)

ankrgyl:
This enables more advanced cors

### DIFF
--- a/packages/proxy/edge/index.ts
+++ b/packages/proxy/edge/index.ts
@@ -75,13 +75,13 @@ export function EdgeProxyV1(opts: ProxyOpts) {
   const whitelist = opts.whitelist || defaultWhitelist;
   return async (request: Request, ctx: EdgeContext) => {
     // If the host is not in the whitelist, return a 403.
-    const origin = request.headers.get("origin");
+    const origin = request.headers.get("Origin");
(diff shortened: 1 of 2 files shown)


packages/proxy/edge/index.ts
```diff
@@ -64,9 +72,29 @@ async function handleOptions(request: Request) {
...
-      return handleOptions(request);
+    // If the host is not in the whitelist, return a 403.
+    const origin = request.headers.get("origin");
+    if (
+      opts.cors &&
+      (!origin ||
+        !whitelist.some(
+          (w) =>
+            w === origin ||
+            (typeof w !== "string" && w instanceof RegExp && w.test(origin)),
```
manugoyal, line 85: Why do we need `typeof w !== string` if we have `w instanceof RegExp`?

manugoyal (approved):
assuming this works. It might be good to add docs for these whitelist env vars to the docker files: https://github.com/braintrustdata/braintrust-deployment/blob/main/docker/docker-compose.full.yml

packages/proxy/edge/index.ts
ankrgyl: We do not

PR #41, opened 2024-04-02
Convert image urls to base64 for claude models

aphinx:
Claude doesn't natively support image URLs, so for convenience we should download valid image URLs and convert them to base64 format.

### DIFF
--- a/packages/proxy/src/providers/anthropic.ts
+++ b/packages/proxy/src/providers/anthropic.ts
@@ -166,33 +166,10 @@ const allowedImageTypes = [
 ];
 
 function arrayBufferToBase64(buf: ArrayBuffer): string {
-  const base64Chars =
-    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
   const bytes = new Uint8Array(buf);
-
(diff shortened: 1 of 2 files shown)


packages/proxy/src/providers/anthropic.ts
```diff
@@ -157,13 +157,96 @@ function makeAnthropicImageBlock(base64Image: string): AnthropicImageBlock {
...
-export function openAIContentToAnthropicContent(
+const maxImageBytes = 5 * 1024 * 1024;
+const allowedImageTypes = [
+  "image/jpeg",
+  "image/png",
+  "image/gif",
+  "image/webp",
+];
+
+function arrayBufferToBase64(buf: ArrayBuffer): string {
```
ankrgyl, line 168: let's try to find something (either through node compat, or `btoa`) that lets us use an existing implementation

ankrgyl (commented):
Awesome!!

packages/proxy/src/providers/anthropic.ts
aphinx: `btoa` did the trick

ankrgyl (approved):

Issue #44, opened 2024-04-06
Encrypt vector embeddings


morgante:
In my reading of the source code, it looks like vector embeddings are not currently encrypted. This is potentially problematic, since they are (somewhat) reversible.

Would you be open to a pull request to encrypt content using the input, so it's only retrievable when you have access to the original source text?

ankrgyl:
Hi @morgante! Thanks for the report. Yes, definitely open to a PR. I'm not totally clear on where you'd want them to be encrypted that they are not (e.g. the results cache should cache the ouput), but a PR would help clarify!

morgante:
Great. Specifically, I don't want the embedding to be retrievable unless you have the original input chunk. I'll put together a PR.

ankrgyl:
Ok got it! I believe that to currently be the case, but I am likely missing something you're observing. Looking forward to the PR.

PR #46, opened 2024-04-08
Ensure cache encryption is not reversible

morgante:
Fixes https://github.com/braintrustdata/braintrust-proxy/issues/44

As demonstrated in the test script, the previous encryption strategy was reversible as long as you had the OpenAI key used. At least for my use case, this violates the guarantees we try to give customers of not retaining copies of their data.

This switches it to use the request data itself as the encryption key, such that you can't decrypt the cache unless you actually have the original user request in hand.

Note: I don't have a Vercel or Node server handy, so this has only been tested with Wrangler/Cloudflare.


apis/cloudflare/src/encrypt.ts
```diff
@@ -1,67 +0,0 @@
-function base64ToArrayBuffer(base64: string) {
```
ankrgyl, line 1: nice :)

packages/proxy/test/decrypt.ts
```diff
@@ -0,0 +1,21 @@
+import { decryptMessage } from "../src/encrypt";
```
ankrgyl, line 1: I think we should either remove this file or introduce `vitest`/`jest` into the repo and make this a test. Happy to help with the latter if you're open to it, or I can clean it up in a follow up change.

packages/proxy/src/proxy.ts
```diff
@@ -172,19 +172,22 @@ export async function proxyV1({
   const orgName = proxyHeaders[ORG_NAME_HEADER];
   const endpointName = proxyHeaders[ENDPOINT_NAME_HEADER];
 
-  const cacheKey =
-    "aiproxy/proxy/v1:" +
-    (await digest(
+  // Data key is computed from the input data and used for both the cache key and as an input to the encryption key.
+  const dataKey = (await digest(
```
ankrgyl, line 176: This logic looks pretty straightforward to me.

packages/proxy/src/proxy.ts
```diff
@@ -172,19 +172,22 @@ export async function proxyV1({
...
       JSON.stringify({
         url,
         body,
         authToken: cacheKeyOptions.excludeAuthToken || authToken,
         orgName: cacheKeyOptions.excludeOrgName || orgName,
         endpointName,
-      }),
+      })
```
ankrgyl, line 183: Did you run the pre-commit hooks? You need to install https://pre-commit.com/ and then just run

```
pre-commit 
```

ankrgyl (commented):
Thank you immensely for contributing this! I see what you mean about wanting to encrypt using the body as well. I will run our internal tests on this and report if there are any issues.

packages/proxy/test/decrypt.ts
morgante: Ok, I removed it. Feel free to add as a test later if you decide to incorporate vitest.

ankrgyl (approved):
Thank you! This is looking good. We need to resolve some stuff with our internal tests, because we cache a bunch of things. Once we figure this out, I'll land this and deploy!

Let me know if you are blocked on this, and if so, we'll make sure to move things along faster.

packages/proxy/src/proxy.ts
morgante: I didn't see a format or prettier command, but ran pre-commit now.

morgante:
> Let me know if you are blocked on this, and if so, we'll make sure to move things along faster.

Not blocked, I've already deployed a version of this for our usage.

ankrgyl:
Ok we are ready to land this! Thanks again for the contribution

PR #59, opened 2024-05-27
Fix caching to store binary data

ankrgyl:
If the data is compressed, the way we store cached data only sometimes works. This change updates it to be base64 encoded, and uses that (if present).

We can eventually remove the `text` case, but I didn't want to break people's existing cached data.

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -43,6 +43,7 @@ type CachedData = {
   headers: Record<string, string>;
 } & (
   | {
+      // DEPRECATION_NOTICE: This can be removed in a couple weeks since writing (e.g. June 9 2024 onwards)
       body: string;
     }
   | {
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -16,6 +16,7 @@ import {
 } from "@schema";
 import {
   flattenChunks,
+  flattenChunksArray,
   getRandomInt,
   isEmpty,
   isObject,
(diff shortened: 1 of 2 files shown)


manugoyal (approved):

packages/proxy/src/proxy.ts
```diff
@@ -300,11 +317,12 @@ export async function proxyV1({
           controller.enqueue(chunk);
         },
         async flush(controller) {
-          const text = flattenChunks(allChunks);
+          const data = flattenChunksArray(allChunks);
+          const dataB64 = btoa(String.fromCharCode(...data));
```
morgante, line 321: @ankrgyl `btoa` should be avoided in production where possible.

ankrgyl: Why? Is there a known set of vulnerabilities or similar?

manugoyal: It seems they are deprecated in nodejs: https://stackoverflow.com/questions/68849233/convert-a-string-to-base64-in-javascript-btoa-and-atob-are-deprecated. Apparently the `Buffer` object is the recommended way to do these binary conversions.

ankrgyl: Sure but it isn't in Cloudflare / workerd. I'll look into turning nodejs compat on, but I'm not sure there's any clear technical benefit within `workerd` (and perhaps some disadvantages, eg bundle size).

morgante: I haven't checked in Cloudflare, but our Node app had stack overflows with `btoa` that was resolved by switching to `Buffer`. String.fromCharCode should also be avoided, since it had some problems with non-ASCII data - for worked, I believe `TextEncoder` is preferred.

ankrgyl: https://github.com/braintrustdata/braintrust-proxy/pull/60

Issue #63, opened 2024-06-07
Something err when stream is not true in requst body


GuanJdoJ:
I run a local service according to this [document](https://github.com/braintrustdata/braintrust-proxy/blob/main/apis/node/README.md).  

It is OK when `"stream": true` is set in request body. But if you don't set the stream, it will fail. 

Success:
``` curl
curl --location 'http://localhost:8001/proxy/v1/chat/completions' \
--header 'Authorization: Bearer $BRAINTRUST_API_KEY' \
--header 'Content-Type: application/json' \
--data '{
    "model": "gpt-3.5-turbo",
    "messages": [
        {
            "role": "user",
            "content": "What is your model?"
        }
    ],
    "stream": true,
    "max_tokens": 150,
    "seed": 1
}'
```

Failed:
``` curl
curl --location 'http://localhost:8001/proxy/v1/chat/completions' \
--header 'Authorization: $BRAINTRUST_API_KEY' \
--header 'Content-Type: application/json' \
--data '{
    "model": "gpt-3.5-turbo",
    "messages": [
        {
            "role": "user",
            "content": "What is your model?"
        }
    ],
    "max_tokens": 150,
    "seed": 1
}'
```

But call `https://braintrustproxy.com` will be OK. Why?
```curl
curl --location 'https://braintrustproxy.com/proxy/v1/chat/completions' \
  -H "Content-Type: application/json" \
  -d '{
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "user",
        "content": "What is a proxy?"
      }
    ],
    "seed": 1
  }' \
  -H "Authorization: Bearer $OPENAI_API_KEY"
```

ankrgyl:
What is the error that you see?

GuanJdoJ:
No specific error reported. I use postman. It shows "Error: Decompression failed" in console.

ankrgyl:
Got it. It may be something related to a compression header being forwarded incorrectly. If you need a short term workaround, I would recommend running the Cloudflare proxy (apis/cloudflare) locally (via `npx wrangler dev`). It's more tested than the node one.

GuanJdoJ:
I will try cloudflare. But I still want to implement it in node. I want to make sure that both the openai eventstream and json results can be processed in `@braintrust/proxy`. I would really appreciate your support.😊

GuanJdoJ:
[...]

PR #65, opened 2024-06-09
fix: fix decompression failed in nodejs

GuanJdoJ:
#63


ankrgyl:
I am working on a change that fixes this, among a few other things, more comprehensively. The change you made is close, but does not work for certain platforms (e.g. workerd) which do not decompress `fetch` results automatically.

Here's the fix: https://github.com/braintrustdata/braintrust-proxy/pull/68/files#diff-f9467ab4882d85829b5ae101d098958cb1f6dafcb726ed6baee1dc26a5e34e9bR300

If you want to port this logic into an independent change here, happy to review that, otherwise we will land the larger change in the next few days.

ankrgyl:
Going to close this out since this functionality has been upstreamed.

PR #68, opened 2024-06-17
Add tracing, factor out some helper functions, fix a few bugs with Nodejs mode

ankrgyl:
This change is a bunch of small tweaks. To summarize:

* Expose a few functions (like app url) out so we can access them when using it as a library
* Fix a series of streaming/incompat bugs that have accumulated between cloudflare's workerd runtime and node. For example, Node's `fetch` decompresses data, while workerd's does not. Also, we weren't properly closing streams based on how we use Vercel's AI SDK parser, so I just rewrote that to use our own streaming logic and pulled out the parser code.
* Add support for tracing.


packages/proxy/src/providers/openai.ts
```diff
@@ -0,0 +1,249 @@
...
+interface ChatCompletionChunk {
+  id: string;
+  choices: Array<ChatCompletionChunkChoice>;
+  created: number;
+  model: string;
+  object: string;
+}
+
+// https://github.com/openai/openai-node/blob/07b3504e1c40fd929f4aae1651b83afc19e3baf8/src/resources/chat/completions.ts#L43-L49
+// Updated for https://github.com/openai/openai-node/commit/f10c757d831d90407ba47b4659d9cd34b1a35b1d
```
manugoyal, line 16: Is this comment superseded by the next line?

packages/proxy/src/proxy.ts
```diff
@@ -65,6 +71,12 @@ export interface CacheKeyOptions {
   excludeOrgName?: boolean;
 }
 
+export interface SpanLogging {
```
manugoyal, line 74: nit: for uniformity, should we name this type `SpanLogger`?

packages/proxy/src/proxy.ts
```diff
@@ -65,6 +71,12 @@ export interface CacheKeyOptions {
...
+  setName: (name: string) => void;
+  log: (args: ExperimentLogPartialArgs) => void;
+  end: () => void;
+}
```
manugoyal, line 78: do we need to explicitly flush the logger here, or is it guaranteed that the `beforeExit` event will fire in all environments that we run this?

manugoyal (approved):

packages/proxy/src/providers/openai.ts
ankrgyl: Not sure. It's just copy/pasted from Vercel's code. Would rather not change it.

packages/proxy/src/proxy.ts
ankrgyl: yep

packages/proxy/src/proxy.ts
ankrgyl: I think it's ok to make it the responsibility of the caller (which is currently our own tightly controlled server, where `asyncFlush` is false by default anyway).

PR #69, opened 2024-06-20
Update model definitions

tara-nagar:
Clean up the model definition and adding bedrock definition to the new Claude model.


ankrgyl (approved):
🙏

PR #77, opened 2024-07-05
Re-add - Update model and secret types

tara-nagar:
Re-applying changes from #75 that were reverted


ankrgyl (approved):
Let's hold off on landing until the corresponding changes in the main repo are ready

PR #85, opened 2024-07-25
Change token cost to per million

tara-nagar:
Updating the model token costs to per million to hopefully avoid any weirdness with rounding floating point decimals. This is also how we display the costs in the UI and its how token costs are most commonly reported.


ankrgyl:
Should we sketch out the migration too?

PR #87, opened 2024-07-30
Check crypto

ankrgyl:


packages/proxy/src/encrypt.ts
```diff
@@ -30,7 +46,9 @@ export async function decryptMessage(
   keyString: string,
   iv: string,
   message: string,
-): Promise<string> {
+): Promise<string | undefined> {
+  if (!isCryptoAvailable()) return undefined;
```
manugoyal, line 50: nit: it would be nice to use `getSubtleCrypto` function instead of raw `crypto.sutble`, so that the `isCryptoAvailable` function is checking exactly the same thing we're using.

manugoyal (approved):

packages/proxy/src/encrypt.ts
ankrgyl: fixed

PR #94, opened 2024-09-01
Allow encoding the org name in the URL

ankrgyl:
This way, you can create a base url that ends with `/btorg/org name` and utilize `OPENAI_BASE_URL` to force a specific org.

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -182,7 +182,7 @@ export async function proxyV1({
     .filter((p) => p.trim() !== "")
     .map((d) => decodeURIComponent(d));
 
-  if (pieces.length > 2 && pieces[0] === "btorg") {
+  if (pieces.length > 2 && pieces[0].toLowerCase() === "btorg") {
     orgName = pieces[1];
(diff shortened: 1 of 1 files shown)


packages/proxy/src/proxy.ts
```diff
@@ -175,6 +175,18 @@ export async function proxyV1({
     proxyHeaders[FORMAT_HEADER],
   );
 
+  let orgName = proxyHeaders[ORG_NAME_HEADER];
+
+  const pieces = url
+    .split("/")
+    .filter((p) => p.trim() !== "")
+    .map((d) => decodeURIComponent(d));
```
manugoyal, line 183: should we lowercase everything too?

manugoyal (approved):

packages/proxy/src/proxy.ts
ankrgyl: org names are case sensitive?

manugoyal: True. I was thinking about the `btorg` part, maybe we just lowercase that comparison.

ankrgyl: done

PR #110, opened 2024-10-29
Proxy & log speech-to-speech sessions on OpenAI Realtime API.

kevin1:


apis/cloudflare/src/realtime-logger.ts
```diff
@@ -0,0 +1,772 @@
...
+// - Where the OpenAI implementation differs from their type spec.
+// - Replace the fields we don't use with `z.any()` for more permissive parsing.
+const baseMessageSchema = z.object({
+  event_id: z.string(),
+});
+
+const audioFormatTypeSchema = z.enum(["pcm16", "g711_ulaw", "g711_alaw"]);
+type AudioFormatType = z.infer<typeof audioFormatTypeSchema>;
+
+const turnDetectionServerVadTypeSchema = z.object({
```
ankrgyl, line 19: small nit: split the schema definitions into a separate file

apis/cloudflare/src/realtime-logger.ts
```diff
@@ -0,0 +1,772 @@
...
+        algorithm: "a",
+        sample_rate: 8000,
+      };
+    default:
+      const x: never = audioFormat;
+      throw new Error(`Unknown audio format ${JSON.stringify(x)}`);
+  }
+}
+
+/**
```
ankrgyl, line 464: i wonder if we shuold move this into the SDK (perhaps in a separate change). So you can use it outside of the proxy too 🤔

ankrgyl (approved):

apis/cloudflare/src/realtime-logger.ts
kevin1: Done

apis/cloudflare/src/realtime-logger.ts
kevin1: Will do in a separate change. Probably also needs some API cleanup before we can commit to maintaining it as part of the SDK.

PR #114, opened 2024-11-02
Accept temporary credentials for chat completions API.

kevin1:


packages/proxy/src/proxy.ts
```diff
@@ -370,9 +375,38 @@ export async function proxyV1({
...
         async (model) => {
+          // First, try to use temp credentials, because then we'll get access
+          // to the model.
+          let cachedAuthToken: string | undefined;
+          if (
+            useCredentialsCacheMode !== "never" &&
+            isTempCredential(authToken)
+          ) {
+            const { credentialCacheValue, jwtPayload } =
+              await verifyTempCredentials({
```
ankrgyl, line 386: if the verification fails should we fall back to normal credential verification? Eg if it's a valid JWT by the control plane's standards and gets accidentally slipped through as a temp credential?

ankrgyl (approved):

packages/proxy/src/proxy.ts
kevin1: I think because this is gated by isTempCredential, we would expect normal verification to fail. We generally can’t trust the inputs in this function because it could be called directly without going through the control plane.

PR #126, opened 2024-12-19
Add provider as response header - BRA-1882

tara-nagar:
Adds the provider used to complete an llm call as a response header

Easier to see changes with whitespace hidden


packages/proxy/src/proxy.ts
```diff
@@ -449,6 +453,10 @@ export async function proxyV1({
       }
       proxyResponseHeaders[name] = value;
     });
+    if (provider) {
+      setHeader(PROVIDER_HEADER, provider);
+      proxyResponseHeaders[PROVIDER_HEADER] = provider;
+    }
```
ankrgyl, line 459: i wonder if we should use the secret name, so that if you have eg multiple openai endponts, we can distinguish?

tara-nagar: oh yeah that's the value it's actually using, i'll name it better

sachinpad (commented):
Maybe add a small test?

packages/proxy/src/proxy.ts
```diff
@@ -83,6 +83,8 @@ export const FORMAT_HEADER = "x-bt-stream-fmt";
 export const LEGACY_CACHED_HEADER = "x-cached";
 export const CACHED_HEADER = "x-bt-cached";
 
+export const PROVIDER_HEADER = "x-bt-provider";
```
ankrgyl, line 86: hmm actually now that i think about it, since `x-bt-endpoint-name` lets you filter to select this, maybe we should name it `x-bt-used-provider`?

ankrgyl: sorry this was a typo on my part. I meant `x-bt-used-endpoint`

ankrgyl (approved):

PR #127, opened 2024-12-20
Add reasoning effort (and set a pattern for more o1 customization)

ankrgyl:


packages/proxy/src/proxy.ts
```diff
@@ -1024,6 +1024,7 @@ async function fetchOpenAI(
 
   // TODO: Ideally this is encapsulated as some advanced per-model config
   // or mapping, but for now, let's just map it manually.
+  console.log("bodyData", bodyData);
```
tara-nagar, line 1027: did you mean to leave this in?

tara-nagar (approved):

packages/proxy/src/proxy.ts
ankrgyl: I did not :)

PR #128, opened 2024-12-26
[bra-1870] make the body (json) deterministic in the cached key

ibolmo:
This ensures caches are more consistent even if the params change order


packages/proxy/src/proxy.ts
```diff
@@ -1,26 +1,39 @@
...
+  ModelFormat,
+  translateParams,
+} from "@schema";
 import {
   createParser,
   type EventSourceParser,
   type ParsedEvent,
   type ReconnectInterval,
 } from "eventsource-parser";
+import stringify from "json-stringify-deterministic";
```
ankrgyl, line 19: could you import this as something more specific, like `stringifyDeterministic`

.tool-versions
```diff
@@ -0,0 +1 @@
+pnpm 8.15.5
```
ankrgyl, line 1: Hmm why do you need this? The lockfile specification in `pnpm-lock` should be enough right?

ibolmo: Yeah, it's a bit tricky. 

pnpm is at 9.x right now (and actually 10.x coming soon), so we're behind on pnpm upgrades. my global pnpm is 9.x and when I opened this up in a separate cursor instance when I did pnpm install/add it updated the lock file to 9.x version. 

This file will both give folks a notice that we're still on 8.x and two that if they're using [asdf](https://asdf-vm.com/) that will be smart to ensure that we run pnpm commands on 8.x.

Likely not needed if we were on the 9.x across our repos. It's on my TODO :D

packages/proxy/src/proxy.ts
```diff
@@ -766,7 +766,7 @@ async function fetchModelLoop(
 
     const modelSpec =
       (model !== null
-        ? (secret.metadata?.customModels?.[model] ?? AvailableModels[model])
+        ? secret.metadata?.customModels?.[model] ?? AvailableModels[model]
```
ibolmo, line 769: I'm not yet sure why in CI this change was necessary but locally (with presumably the same prettier version) it prefers with `()`

PR #142, opened 2025-02-04
support anthropic structured outputs

sachinpad:


packages/proxy/src/providers/anthropic.ts
```diff
@@ -332,6 +341,7 @@ export function anthropicEventToOpenAIEvent(
 export function anthropicCompletionToOpenAICompletion(
   completion: AnthropicCompletion,
   isFunction: boolean,
+  isStructuredOutput: boolean,
 ): ChatCompletion {
   console.log("COMPLETION", JSON.stringify(completion, null, 2));
```
ankrgyl, line 346: mind removing this console.log statement

ankrgyl (commented):
Can we apply this technique to OpenAI flavored LLMs that do not have structured output (e.g. llama hosted on together)

packages/proxy/src/providers/anthropic.ts
sachinpad: Done

sachinpad:
> Can we apply this technique to OpenAI flavored LLMs that do not have structured output (e.g. llama hosted on together)

Done

package.json
```diff
@@ -21,5 +21,8 @@
     "vite-tsconfig-paths": "^4.3.2",
     "vitest": "^2.1.3"
   },
-  "packageManager": "pnpm@8.15.5"
+  "packageManager": "pnpm@8.15.5",
+  "resolutions": {
+    "zod": "3.22.4"
+  }
```
ankrgyl, line 27: Hm I believe we've moved away from doing this (@manugoyal @ibolmo have context)

packages/proxy/src/providers/bedrock.ts
```diff
@@ -341,14 +350,48 @@ export async function fetchConverse({
...
+    })
+    .safeParse(oaiParams.response_format);
+  if (parsed.success) {
+    isStructuredOutput = true;
+    if (toolConfig) {
+      throw new ProxyBadRequestError(
+        "Structured output is not supported with tools",
+      );
+    }
+    toolConfig = {
```
ankrgyl, line 370: don't you need to delete the `response_format` param? Maybe converse doesn't barf?

package.json
sachinpad: In this PR's case, the version of zod was different between the proxy and SDK, which was causing some strange type errors. I guess we could manually ensure they are the same?

packages/proxy/src/providers/bedrock.ts
```diff
@@ -250,11 +259,11 @@ function translateTools(tools: ChatCompletionTool[]): ToolConfiguration {
         name: tool.function.name,
         description: tool.function.description,
         inputSchema: {
-          json: tool.function.parameters,
+          json: tool.function.parameters as any,
```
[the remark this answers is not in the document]
sachinpad: Looks like the proxy repo doesn't have as strict of a linter. It's needed since OpenAI and amazon define the json schema type a bit differently and typescript isn't smart enough to see that they're actually the same

packages/proxy/src/providers/bedrock.ts
sachinpad: yes, will do, although looks like it doesn't go crazy

package.json
ankrgyl: I think the way we accomplished this in the past was by pinning it in `pnpm-lock.yaml` instead. @manugoyal or @ibolmo can help you do this the right way.

packages/proxy/src/providers/bedrock.ts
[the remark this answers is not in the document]
ankrgyl: got it. mind leaving a comment?

package.json
manugoyal: Will defer to @ibolmo, as I was only familiar with the `resolutions` approach, but now see we are using overrides in other parts of the codebase, such as our internal repo. A quick google suggests that `pnpm.overrides` is preferred for pnpm projects.

packages/proxy/src/providers/bedrock.ts
sachinpad: Oh, just realized we don't use oaiParams anymore. It's translated, not modified

packages/proxy/src/providers/bedrock.ts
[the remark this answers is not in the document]
sachinpad: Done

package.json
ibolmo: Yeah I did an `override`: https://github.com/braintrustdata/braintrust-sdk/pull/469/files

ankrgyl (approved):

PR #144, opened 2025-02-05
added new gemini models

Abarbir12:


packages/proxy/schema/models.ts
```diff
@@ -1000,11 +1000,27 @@ export const AvailableModels: { [name: string]: ModelSpec } = {
     multimodal: true,
   },
   "gemini-2.0-flash-exp": {
```
richrliu, line 1002: Can we add `gemini-2.0-flash-latest` with the same params?

packages/proxy/schema/models.ts
```diff
@@ -1000,11 +1000,27 @@ export const AvailableModels: { [name: string]: ModelSpec } = {
...
+    format: "google",
+    flavor: "chat",
+    input_cost_per_mil_tokens: 0.1, // TODO: Appears to be free for now?
+    output_cost_per_mil_tokens: 0.4,
+    displayName: "Gemini 2.0 Flash Exp",
```
richrliu, line 1007: Can we drop the "Exp" from the name - I think it's now GA and no longer experimental

richrliu (commented):
Thanks for adding!

ankrgyl (approved):

PR #146, opened 2025-02-06
add gemini-2.0-flash-001

richrliu:
from the [gemini api docs](https://ai.google.dev/gemini-api/docs/models/gemini#gemini-2.0-flash) it seems like the alias `gemini-2.0-flash-latest` is not yet active. Need to use `gemini-2.0-flash-001` instead.


Abarbir12:
Hey richard, we have a change that will go live soon resolving this. I'll keep you updated and lyk when it's live. Thanks for pointing it out

richrliu:
noted, thank you! will close this PR

PR #149, opened 2025-02-08
support tool calls and structured outputs for gemini

sachinpad:


packages/proxy/src/proxy.ts
```diff
@@ -1522,6 +1521,111 @@ async function fetchAnthropic(
   };
 }
 
+function pruneSchema(schema: any): any {
```
ankrgyl, line 1524: If this is specific to google, can you either name it as such or move it into the google-specific file?

packages/proxy/src/proxy.ts
```diff
@@ -1522,6 +1521,111 @@ async function fetchAnthropic(
...
+      result[key] = pruneSchema(value);
+    } else {
+      result[key] = value;
+    }
+  }
+
+  return result;
+}
+
+function googleTools(params: ChatCompletionCreateParams) {
```
ankrgyl, line 1565: nit: maybe rename `openAIToolsToGoogleTools`?

packages/proxy/src/proxy.ts
sachinpad: Done

packages/proxy/src/proxy.ts
sachinpad: Done

ankrgyl (approved):

PR #159, opened 2025-02-17
Implement cache-control directives for max-age, no-cache, and no-store

choochootrain:
## Requests
Requests can specify cache behavior and TTL using a limited set of [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) directives. The following request directives are implemented:
* `no-cache`,`no-store` (these are treated as equivalent)
* `max-age` (up to 90 days)

The `x-bt-cache-ttl` header is also added to request a specific TTL (between 1 second and 90 days)

Note that setting the Cache-Control header will override the behavior of the `x-bt-use-cache` header.  

### Examples:
* Request with `x-bt-cache-ttl: 604800` is equivalent to `x-bt-use-cache: auto`
* Request with `Cache-control: no-cache` is equivalent to `x-bt-use-cache: never`
* Request with `Cache-control: no-store` is equivalent to `x-bt-use-cache: never`
* Request with `Cache-control: max-age=0` is equivalent to `x-bt-use-cache: never`
* Request with `x-bt-cache-ttl: 77600` will cache requests on supported paths where temperature is 0 or a seed is set for 90 days
* Request with `x-bt-cache-ttl: 77600`, `Cache-control: max-age=60` will cache all requests on supported paths for 90 days but will only use already cached results <= 60 seconds old
* Request with `x-bt-cache-ttl: 77600`, `x-bt-use-cache: always` will cache all requests on supported paths for 90 days

### Unimplemented request directives
  - `max-stale`, `stale-if-error`: Might be useful to implement if the origin API is down
  - `min-fresh`, `no-transform`, `only-if-cached`: Not relevant for a caching proxy

## Responses
The proxy will respond with basic information about the age of cached data using the `max-age` directive and the [Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Age) header. If the request was not placed into the cache, these values will not be in the response.

### Examples:
* Response with `x-bt-cached: MISS`, `Cache-control: max-age=776000`, and `Age: 0` missed the cache and was just retrieved from the provider and cached for 90 days
* Response with `x-bt-cached: MISS` missed the cache and was not cached
* Response with `x-bt-cached: HIT`, `Cache-control: max-age=60`, and `Age: 15` was retrieved from the cache and will be cached for another 45 seconds

### Unimplemented response directives
  - `no-cache`, `no-store`, `must-revalidate`, `proxy-revalidate`, `must-understand`, `private`, `public`, `stale-while-revalidate`, `stale-if-error`: Proxy is agnostic to how consumers handle data
  - `s-maxage`: Ignored since proxy doesn't support shared caching
  - `immutable`: Maybe respond with this directive if the proxy keeps the always/auto/never cache semantics
  
- [ ] Followup PR to refactor `timestamp` into non-optional property of cached data once cache is cycled

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -23,6 +23,7 @@ import {
   isEmpty,
   isObject,
   parseAuthHeader,
+  parseNumericHeader,
 } from "./util";
 import {
   anthropicCompletionToOpenAICompletion,
@@ -71,7 +72,6 @@ import { openAIChatCompletionToChatEvent } from "./providers/openai";
 import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
 import { importPKCS8, SignJWT } from "jose";
 import { z } from "zod";
-import { CANCELLED } from "node:dns";
 
 type CachedData = {
   headers: Record<string, string>;
@@ -87,13 +87,11 @@ type CachedData = {
     }
 );
 
-export const CACHE_CONTROL_HEADER = "cache-control";
-export const AGE_HEADER = "age";
 // XXX determine appropriate max cache TTL
-export const MAX_CACHE_TTL = 90 * 24 * 60 * 60; // 90 days
-export const DEFAULT_CACHE_TTL = 7 * 24 * 60 * 60; // 7 days
-// XXX maybe rename this if standard cache control becomes default behavior
+const MAX_CACHE_TTL = 90 * 24 * 60 * 60; // 90 days
+const DEFAULT_CACHE_TTL = 7 * 24 * 60 * 60; // 7 days
 export const CACHE_HEADER = "x-bt-use-cache";
+export const CACHE_TTL_HEADER = "x-bt-cache-ttl";
 export const CREDS_CACHE_HEADER = "x-bt-use-creds-cache";
 export const ORG_NAME_HEADER = "x-bt-org-name";
 export const ENDPOINT_NAME_HEADER = "x-bt-endpoint-name";
@@ -187,7 +185,7 @@ export async function proxyV1({
           h === "priority" ||
           h === "referer" ||
           h === "user-agent" ||
-          h === CACHE_CONTROL_HEADER
+          h === "cache-control"
         ),
     ),
   );
@@ -207,19 +205,19 @@ export async function proxyV1({
 
   let useCacheMode = legacyCacheMode;
   const cacheControlDirectives = cacheControlParse(
-    proxyHeaders[CACHE_CONTROL_HEADER] || "",
+    proxyHeaders["cache-control"] || "",
   );
   // XXX does it make sense to have separate cache control for read and write?
   if (
     cacheControlDirectives?.["no-cache"] ||
-    cacheControlDirectives?.["no-store"]
+    cacheControlDirectives?.["no-store"] ||
+    cacheControlDirectives?.["max-age"] === 0
   ) {
     useCacheMode = "never";
   }
-  const cacheTTL = Math.min(
-    Math.max(1, cacheControlDirectives?.["max-age"] ?? DEFAULT_CACHE_TTL),
-    DEFAULT_CACHE_TTL,
-  );
+  const requestedTTL =
+    parseNumericHeader(proxyHeaders, CACHE_TTL_HEADER) ?? DEFAULT_CACHE_TTL;
+  const cacheTTL = Math.min(Math.max(1, requestedTTL), MAX_CACHE_TTL);
 
   const useCredentialsCacheMode = parseEnumHeader(
     CACHE_HEADER,
@@ -346,11 +344,11 @@ export async function proxyV1({
       }
       setHeader(LEGACY_CACHED_HEADER, "true");
       setHeader(CACHED_HEADER, "HIT");
-      setHeader(CACHE_CONTROL_HEADER, `max-age=${cacheTTL}`);
+      setHeader("cache-control", `max-age=${cacheTTL}`);
       // XXX simplify once all cached data has timestamp
       if (cachedData.timestamp) {
         setHeader(
-          AGE_HEADER,
+          "age",
           `${Math.floor(getCurrentUnixTimestamp() - cachedData.timestamp)}`,
         );
       }
@@ -528,8 +526,8 @@ export async function proxyV1({
     setHeader(LEGACY_CACHED_HEADER, "false"); // We're moving to x-bt-cached
     setHeader(CACHED_HEADER, "MISS");
     if (useCache) {
-      setHeader(CACHE_CONTROL_HEADER, `max-age=${cacheTTL}`);
-      setHeader(AGE_HEADER, "0");
+      setHeader("cache-control", `max-age=${cacheTTL}`);
+      setHeader("age", "0");
     }
 
     if (stream && proxyResponse.ok && useCache) {
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -87,7 +87,6 @@ type CachedData = {
     }
 );
 
-// XXX determine appropriate max cache TTL
 const MAX_CACHE_TTL = 90 * 24 * 60 * 60; // 90 days
 const DEFAULT_CACHE_TTL = 7 * 24 * 60 * 60; // 7 days
 export const CACHE_HEADER = "x-bt-use-cache";
@@ -97,7 +96,6 @@ export const ORG_NAME_HEADER = "x-bt-org-name";
 export const ENDPOINT_NAME_HEADER = "x-bt-endpoint-name";
 export const FORMAT_HEADER = "x-bt-stream-fmt";
 
-export const LEGACY_CACHED_HEADER = "x-cached";
 export const CACHED_HEADER = "x-bt-cached";
 
 export const USED_ENDPOINT_HEADER = "x-bt-used-endpoint";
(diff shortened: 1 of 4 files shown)


packages/proxy/src/proxy.ts
```diff
@@ -70,9 +71,12 @@ import { openAIChatCompletionToChatEvent } from "./providers/openai";
 import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
 import { importPKCS8, SignJWT } from "jose";
 import { z } from "zod";
+import { CANCELLED } from "node:dns";
```
sachinpad, line 74: unused?

packages/proxy/src/proxy.ts
```diff
@@ -83,6 +87,12 @@ type CachedData = {
     }
 );
 
+export const CACHE_CONTROL_HEADER = "cache-control";
+export const AGE_HEADER = "age";
```
sachinpad, line 91: nit: just hardcode since these are well understood?

packages/proxy/src/proxy.ts
```diff
@@ -83,6 +87,12 @@ type CachedData = {
...
+// XXX determine appropriate max cache TTL
+export const MAX_CACHE_TTL = 90 * 24 * 60 * 60; // 90 days
+export const DEFAULT_CACHE_TTL = 7 * 24 * 60 * 60; // 7 days
```
sachinpad, line 94: no need to export

packages/proxy/src/proxy.ts
```diff
@@ -187,11 +198,29 @@ export async function proxyV1({
   }
 
   // Caching is enabled by default, but let the user disable it
-  const useCacheMode = parseEnumHeader(
+  // XXX is there a clean way to specify "auto" caching behavior using only cache-control directives?
```
sachinpad, line 201: i think i like the way you've currently done it, in which we default to "auto" and "always" requires the explicit header

packages/proxy/src/proxy.ts
```diff
@@ -187,11 +198,29 @@ export async function proxyV1({
...
     CACHE_HEADER,
     CACHE_MODES,
     proxyHeaders[CACHE_HEADER],
   );
+
+  let useCacheMode = legacyCacheMode;
+  const cacheControlDirectives = cacheControlParse(
+    proxyHeaders[CACHE_CONTROL_HEADER] || "",
+  );
+  // XXX does it make sense to have separate cache control for read and write?
```
sachinpad, line 212: i think that does make sense, if you'd like to do that in this change

packages/proxy/src/proxy.ts
```diff
@@ -187,11 +198,29 @@ export async function proxyV1({
...
-  const useCacheMode = parseEnumHeader(
+  // XXX is there a clean way to specify "auto" caching behavior using only cache-control directives?
+  const legacyCacheMode = parseEnumHeader(
     CACHE_HEADER,
     CACHE_MODES,
     proxyHeaders[CACHE_HEADER],
   );
+
+  let useCacheMode = legacyCacheMode;
+  const cacheControlDirectives = cacheControlParse(
```
sachinpad, line 209: should we also specifically intercept max-age=0, and take that to mean "don't cache"?

packages/proxy/src/proxy.ts
```diff
@@ -187,11 +198,29 @@ export async function proxyV1({
...
+  // XXX does it make sense to have separate cache control for read and write?
+  if (
+    cacheControlDirectives?.["no-cache"] ||
+    cacheControlDirectives?.["no-store"]
+  ) {
+    useCacheMode = "never";
+  }
+  const cacheTTL = Math.min(
+    Math.max(1, cacheControlDirectives?.["max-age"] ?? DEFAULT_CACHE_TTL),
+    DEFAULT_CACHE_TTL,
```
sachinpad, line 221: do you mean MAX_CACHE_TTL?

sachinpad: also, i wonder if we should have a separate, bt-specific, header that controls the ttl since from the MDN docs, it doesn't seem like max-age controls the ttl on the server, just how fresh of a response is expected by the client

packages/proxy/src/proxy.ts
```diff
@@ -490,6 +527,10 @@ export async function proxyV1({
     }
     setHeader(LEGACY_CACHED_HEADER, "false"); // We're moving to x-bt-cached
     setHeader(CACHED_HEADER, "MISS");
+    if (useCache) {
```
sachinpad, line 530: do we need to check this? if we're in this path, then we necessarily must have missed the cache right?

packages/proxy/src/proxy.ts
```diff
@@ -70,9 +71,12 @@ import { openAIChatCompletionToChatEvent } from "./providers/openai";
 import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
 import { importPKCS8, SignJWT } from "jose";
 import { z } from "zod";
+import { CANCELLED } from "node:dns";
 
 type CachedData = {
   headers: Record<string, string>;
+  // XXX make this a required field once deployed and cache data is cycled for 1 week (previous max cache TTL)
+  timestamp?: number;
```
sachinpad, line 79: can we store the date directly?

sachinpad (commented):
Also, aside from updating the docs with this new stuff, we should also update them to reflect that "auto" also only caches supported endpoints

packages/proxy/src/proxy.ts
sachinpad: oh jk, i see

packages/proxy/src/proxy.ts
sachinpad: up to you, but imo might be more human-readable if we inspect cache entries

packages/proxy/src/proxy.ts
```diff
@@ -309,55 +331,64 @@ export async function proxyV1({
...
+        ? Math.floor(getCurrentUnixTimestamp() - cachedData.timestamp)
+        : DEFAULT_CACHE_TTL;
+
+      if (!cacheMaxAge || age <= cacheMaxAge) {
+        cacheHits.add(1);
+        for (const [name, value] of Object.entries(cachedData.headers)) {
+          setHeader(name, value);
+        }
+        setHeader(CACHED_HEADER, "HIT");
+        setHeader("cache-control", `max-age=${cacheTTL}`);
```
sachinpad, line 346: should we give the ttl that the user gave here, or how long the entry has left to live?

packages/proxy/src/proxy.ts
sachinpad: I guess the only behavior that i would ever want to use that we don't support is forcibly "busting" the cache, where we don't read the current cache entry if any, but always write--up to you on whether we want to support that, we could also just keep this simple for now

packages/proxy/src/proxy.ts
choochootrain: yeah this should be the requested ttl. ill add to the cached metadata and update

packages/proxy/src/proxy.ts
```diff
@@ -276,10 +300,17 @@ export async function proxyV1({
     bodyData.temperature !== 0 &&
     (bodyData.seed === undefined || bodyData.seed === null);
 
-  const useCache =
+  const readFromCache =
```
sachinpad, line 303: should we check cacheMaxAge right here? may be simpler if we just treat it as equivalent to no-cache

packages/proxy/src/proxy.ts
```diff
@@ -305,59 +336,72 @@ export async function proxyV1({
...
   let stream: ReadableStream<Uint8Array> | null = null;
-  if (useCache) {
+  if (readFromCache) {
     const cached = await cacheGet(encryptionKey, cacheKey);
 
     if (cached !== null) {
-      cacheHits.add(1);
       const cachedData: CachedData = JSON.parse(cached);
+      // XXX simplify once all cached data has a timestamp - assume existing data has age of 7 days
+      const ttl = cachedData.metadata?.ttl ?? DEFAULT_CACHE_TTL;
```
sachinpad, line 345: shouldn't we compute this as something like `expirationTime - now`, unless I'm mistaken?

packages/proxy/src/proxy.ts
```diff
@@ -66,13 +68,20 @@ import {
...
   verifyTempCredentials,
 } from "utils";
+import { differenceInSeconds, formatISO, parseISO } from "date-fns";
 import { openAIChatCompletionToChatEvent } from "./providers/openai";
 import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
 import { importPKCS8, SignJWT } from "jose";
 import { z } from "zod";
 
+type CachedMetadata = {
+  cached_at: string;
```
sachinpad, line 78: i was thinking we could just store the Date object itself--I believe its (de)serialization to JSON should Just Work?

packages/proxy/src/proxy.ts
```diff
@@ -66,13 +68,20 @@ import {
...
+  ttl: number;
```
sachinpad, line 79: reflected by another comment but we could alternatively parameterize this as `(cachedAt, expiration)` which may be easier to work with

packages/proxy/src/proxy.ts
choochootrain: yeah we can check it for the max-age=0 scenario here but otherwise we need to reach into the cache to see if it meets the requested max age

packages/proxy/src/proxy.ts
choochootrain: this variable name is misleading - the response directive for max-age is the amount of time the response is fresh from the moment if was generated. ill rename

sachinpad: ahhh i see, i was mistaken. this makes sense. we could probably also then change the parameterization back to cached_at and total ttl? sorry!

sachinpad (approved):

PR #163, opened 2025-02-21
attach all b64 images in the proxy

sachinpad:


ankrgyl (commented):
* We shouldn't be logging in or propagating the SDK code into the proxy (it's also not a dependency in package.json)
* Extend the `SpanLogger` interface that we pass into the proxy instead
* Also, we shouldn't actually edit the payload we send to the downstream model provider (some _only_ support b64, and there will be a perf impact otherwise). Instead, we should just edit the payload we log.

ankrgyl:
Going to do this a different way

PR #169, opened 2025-02-25
deprecate non-profile bedrock models

sachinpad:


packages/proxy/schema/models.ts
```diff
@@ -1992,6 +1992,31 @@ export const AvailableModels: { [name: string]: ModelSpec } = {
     output_cost_per_mil_tokens: 0.6,
     displayName: "Command Light",
   },
+  // Bedrock deprecated.
+  "anthropic.claude-3-5-haiku-20241022-v1:0": {
```
ankrgyl, line 1996: do you need to mark these as `deprecated`?

ornellaaltunyan (approved):

PR #179, opened 2025-03-07
support proxy speaking anthropic and gemini

sachinpad:


packages/proxy/schema/secrets.ts
```diff
@@ -47,7 +47,11 @@ export const VertexMetadataSchema = BaseMetadataSchema.merge(
   z.object({
     project: z.string().min(1, "Project cannot be empty"),
     authType: z.enum(["access_token", "service_account_key"]),
-    api_base: z.string().url().optional(),
+    api_base: z.union([
```
ankrgyl, line 50: is it better (CC @manugoyal ?) to do

```
    api_base: z.union([
      z.string().url(),
      z.string().length(0),
    ]).nullish(),
```

packages/proxy/src/proxy.ts
```diff
@@ -243,18 +248,24 @@ export async function proxyV1({
...
+  const isGoogleUrl = GOOGLE_URL_REGEX.test(url);
+
   const cacheableEndpoint =
     url === "/auto" ||
     url === "/embeddings" ||
     url === "/chat/completions" ||
     url === "/completions" ||
-    url === "/moderations";
+    url === "/moderations" ||
+    url === "/messages" ||
```
ankrgyl, line 259: should we prefix anthropic and google with `/v1/proxy/anthropic` etc to protect ourselves from providers creating overlapping names in the future?

ankrgyl (commented):
nice!

packages/proxy/schema/secrets.ts
sachinpad: Done

packages/proxy/src/proxy.ts
sachinpad: Done

ankrgyl (approved):

PR #194, opened 2025-03-31
handle exception from iterator.next

sachinpad:


packages/proxy/src/providers/bedrock.ts
```diff
@@ -236,7 +236,22 @@ export async function fetchBedrockAnthropic({
     let idx = 0;
     responseStream = new ReadableStream<Uint8Array>({
       async pull(controller) {
-        const { value, done } = await iterator.next();
+        let value, done;
+        try {
+          ({ value, done } = await iterator.next());
+        } catch (e) {
+          console.warn("Error from iterator.next():", e);
```
ankrgyl, line 243: can you add the context that this is from bedrock+anthropic into the console message

ankrgyl (approved):

packages/proxy/src/providers/bedrock.ts
sachinpad: Done

ankrgyl (approved):

PR #203, opened 2025-04-12
[chore] Add Gemini 2.5 Pro Preview to list of models

fctb12:
This PR adds Gemini 2.5 Pro Preview (which has higher rate limits than the experimental version) to the Braintrust AI proxy.


ankrgyl (approved):
Thank you!

PR #204, opened 2025-04-14
Support tracing through Cloudflare proxy via header

ankrgyl:


apis/cloudflare/src/tracing.ts
```diff
@@ -0,0 +1,187 @@
...
+    const projectName = header.substring(PROJECT_NAME_PREFIX.length);
+    return new SpanComponentsV3({
+      object_type: SpanObjectTypeV3.PROJECT_LOGS,
+      compute_object_metadata_args: {
+        project_name: projectName,
+      },
+    });
+  }
+
+  return SpanComponentsV3.fromStr(header);
```
choochootrain, line 132: should we catch the invalid encoding error here and ignore the header if its malformed?

choochootrain (commented):
looks legit to me. most of tracing.ts comes from `@braintrust/local` so we should probably move it to core since you're already pulling it in here

apis/cloudflare/src/tracing.ts
ankrgyl: hmm interesting idea, is that the behavior you'd want?

choochootrain: the error response you added also works, as long as we don't have some uncaught exception here and result in a 500 response.

apis/cloudflare/src/proxy.ts
```diff
@@ -65,31 +72,59 @@ export async function handleProxyV1(
...
+      );
+    }
+    span = startSpan({
+      state: await cachedLogin({
+        appUrl: braintrustAppUrl(env).toString(),
+        headers: request.headers,
+        cache: credentialsCache,
+      }),
+      type: "llm",
+      name: "LLM",
```
choochootrain, line 116: not blocking but i wonder if there's utility in being able to control the name of this span via headers as well. ill leave it to you to since im not too familiar with the practices we use for span names elsewhere and what we typically see customers doing

apis/cloudflare/src/tracing.ts
ankrgyl: It ends up as a nice 400!

choochootrain (approved):

apis/cloudflare/src/proxy.ts
ankrgyl: That is a pretty neat idea. We can always make it configurable if it comes up

apis/cloudflare/src/tracing.ts
```diff
@@ -0,0 +1,152 @@
...
+export function isBase64Image(s: string): boolean {
+  // Avoid unnecessary (slower) pattern matching
+  if (!s.startsWith("data:")) {
+    return false;
+  }
+
+  return base64ImagePattern.test(s);
+}
+// Being as specific as possible about allowable characters and avoiding greedy matching
+// helps avoid catastrophic backtracking: https://github.com/braintrustdata/braintrust/pull/4831
```
ibolmo, line 84: #nit internal link

apis/cloudflare/src/tracing.ts
```diff
@@ -0,0 +1,152 @@
...
+  const parts = s.match(base64ContentTypePattern);
+  if (!parts) {
+    throw new Error("Invalid base64 image");
+  }
+  const mimeType = parts[1];
+  const data = s.slice(`data:${mimeType};base64,`.length);
+  return { mimeType, data: base64ToArrayBuffer(data) };
+}
+
+export async function cachedLogin({
```
ibolmo, line 100: just a note for myself that the above and here are copy/pasta. perhaps we need to move them in a shared package 🤔

ankrgyl: some of it is, but some of it is different (cloudflare has diff js for base64 encoding/decoding, and the caching logic is a bit different)

PR #209, opened 2025-04-18
Support openai + pdfs

ankrgyl:


packages/proxy/src/providers/openai.ts
```diff
@@ -66,3 +73,60 @@ export function makeFakeOpenAIStreamTransformer() {
...
+// https://platform.openai.com/docs/guides/pdf-files?api-mode=chat
+async function normalizeOpenAIContent(
+  content: ChatCompletionContentPart,
+): Promise<ChatCompletionContentPart> {
+  if (typeof content === "string") {
+    return content;
+  }
+  switch (content.type) {
+    case "image_url":
+      console.log("image_url", content.image_url.url);
```
ibolmo, line 109: Likely need to remove this, https://github.com/braintrustdata/braintrust-proxy/pull/209/files#diff-680db4418ead323337b7b031e9eb51c25ec7d09fae027b6a247f2685c819ff5fR82, https://github.com/braintrustdata/braintrust-proxy/pull/209/files#diff-680db4418ead323337b7b031e9eb51c25ec7d09fae027b6a247f2685c819ff5fR118

packages/proxy/src/providers/openai.ts
```diff
@@ -66,3 +73,60 @@ export function makeFakeOpenAIStreamTransformer() {
...
+        return {
+          type: "file",
+          file: {
+            filename:
+              content.image_url.url.split("/").pop()?.split("?")[0] ??
+              "image.pdf",
+            file_data: base64ToUrl(base64),
+          },
+        };
+      }
```
ibolmo, line 128: give up and return the `content.image_url.url`?

packages/proxy/src/providers/openai.ts
```diff
@@ -66,3 +73,60 @@ export function makeFakeOpenAIStreamTransformer() {
...
+          media: content.image_url.url,
+          allowedMediaTypes: ["application/pdf"],
+          maxMediaBytes: 20 * 1024 * 1024,
+        });
+        console.log(base64.data);
+        return {
+          type: "file",
+          file: {
+            filename:
+              content.image_url.url.split("/").pop()?.split("?")[0] ??
```
ibolmo, line 123: _may_ want to use `/[%\w\-. ]+\.pdf/`

i.e.
```
['http://foo.com/bar/example.pdf', 'gs://bucket/file.pdf', 'http://foo.com/bar/?file=example.pdf', 'http://foo.com/bar/#file=exa%20mple.pdf'].map(u => u.match(/[%\w\-. ]+\.pdf/)[0]).map(decodeURIComponent)
```

```
['example.pdf', 'file.pdf', 'example.pdf', 'exa mple.pdf']
```

packages/proxy/src/providers/openai.ts
```diff
@@ -66,3 +73,60 @@ export function makeFakeOpenAIStreamTransformer() {
...
+              "image.pdf",
+            file_data: base64ToUrl(base64),
```
ibolmo, line 125: TIL 🤦🏻

ibolmo: Do we have a config for the model and their feature support? We may want to provide a helpful message if the model they've chosen doesn't support pdf (or do we want to rely on the AI provider response)

> Supported models
> 
> Only models that support both text and image inputs, such as gpt-4o, gpt-4o-mini, or o1, can accept PDF files as input. [Check model features here](https://platform.openai.com/docs/models).

ibolmo (approved):
Probably a little cleanup (logs mostly) but great find

packages/proxy/src/providers/openai.ts
ankrgyl: whoops thank you

packages/proxy/src/providers/openai.ts
ankrgyl: spruced this up + wrote some tests

packages/proxy/src/providers/openai.ts
ankrgyl: i think we can rely on provider errors to start

packages/proxy/src/providers/openai.ts
ankrgyl: technically it was already doign that (by falling to the `default` case) but let's make it explicit

PR #210, opened 2025-04-18
Download local images in the proxy

ankrgyl:
While developing or self hosting, it's useful to have references to local images.

### DIFF
--- a/packages/proxy/src/providers/openai.ts
+++ b/packages/proxy/src/providers/openai.ts
@@ -122,6 +122,21 @@ async function normalizeOpenAIContent(
             file_data: base64ToUrl(base64),
           },
         };
+      } else if (
+        content.image_url.url.startsWith("http://127.0.0.1") ||
+        content.image_url.url.startsWith("http://localhost")
+      ) {
+        const base64 = await convertMediaToBase64({
+          media: content.image_url.url,
+          allowedMediaTypes: null,
+          maxMediaBytes: 20 * 1024 * 1024,
+        });
+        return {
+          type: "image_url",
+          image_url: {
+            url: base64ToUrl(base64),
+          },
+        };
       }
       return content;
     default:
--- a/packages/proxy/src/providers/util.ts
+++ b/packages/proxy/src/providers/util.ts
@@ -26,7 +26,7 @@ async function convertMediaUrl({
   maxMediaBytes,
 }: {
   url: string;
-  allowedMediaTypes: string[];
+  allowedMediaTypes: string[] | null;
   maxMediaBytes: number | null;
 }): Promise<MediaBlock> {
   const response = await fetch(url);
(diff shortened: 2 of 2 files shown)


edenh (approved):

packages/proxy/src/providers/openai.ts
```diff
@@ -122,6 +122,21 @@ async function normalizeOpenAIContent(
             file_data: base64ToUrl(base64),
           },
         };
+      } else if (
+        content.image_url.url.startsWith("http://127.0.0.1") ||
+        content.image_url.url.startsWith("http://localhost")
```
ibolmo, line 127: curious why we needed the distinction 🤔 oh where we passing those directly to openai rather than converting to base64? 

I wonder if we should create helpers like 

image


Something like:

```ts
{
  "name": "base64",
  "base64": function () {
    return function (text, render) {
      return convertMediaToBase64(render(text));
    }
  }
}
```

Then in the prompt:

```
{{#base64}}{{input.file}}{{/base64}}
```

I know a bit techie, but I could see a future prompt builder where we help with snippets like this.

packages/proxy/src/providers/util.ts
```diff
@@ -39,7 +39,10 @@ async function convertMediaUrl({
     throw new Error("Failed to get content type of the media");
   }
   const baseContentType = contentType.split(";")[0].trim();
-  if (!allowedMediaTypes.includes(baseContentType)) {
+  if (
+    allowedMediaTypes !== null &&
+    !allowedMediaTypes.includes(baseContentType)
```
ibolmo, line 44: #nit you can `if (allowedMediaTypes?.includes(baseContentType))`

packages/proxy/src/providers/openai.ts
ibolmo: may be overkill, but maybe we want to do a lookup in case someone is doing something like `http://test.local/foo.pdf`. 

```
import { lookup } from 'dns/promises';
import { URL } from 'url';

async function isLocalhost(url: string, timeout: number = 1000): Promise<boolean> {
  try {
    const { hostname } = new URL(url);
    
    // Check common localhost names directly
    if (hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1') {
      return true;
    }

    const result = await lookup(hostname, {
      timeout,
      family: 0
    });

    return result.address === '127.0.0.1' || 
           result.address === '::1' || 
           result.address.startsWith('127.');
           
  } catch (error) {
    return false;
  }
}
```

May be useful, for example, if someone builds a sidecar to proxy the requests to Google drive.

ibolmo (approved):

packages/proxy/src/providers/util.ts
ankrgyl: I think that would throw an exception if allowed media types is specified and includes the type (which we dont want)

packages/proxy/src/providers/openai.ts
ankrgyl: This is just to help with local testing -- would rather not add a syscall (`lookup`) into the critical path.

PR #214, opened 2025-04-28
Add Anthropic thinking params and responses support

ibolmo:


packages/proxy/package.json
```diff
@@ -82,7 +82,7 @@
     "@anthropic-ai/sdk": "^0.39.0",
     "@apidevtools/json-schema-ref-parser": "^11.9.1",
     "@aws-sdk/client-bedrock-runtime": "^3.738.0",
-    "@braintrust/core": "^0.0.85",
+    "@braintrust/core": "workspace:*",
```
ankrgyl, line 85: this won't work for the proxy repo on its own, unfortunately. if you made `core` changes, then we'll need to land those, publish them, and then increment this independently.

packages/proxy/schema/index.ts
```diff
@@ -49,10 +49,79 @@ export const modelParamToModelParam: {
   stream_options: null,
   parallel_tool_calls: null,
   response_format: null,
-  reasoning_effort: null,
+  reasoning_effort: "reasoning_effort",
   stop: null,
 };
 
+const effortToBudgetMultiplier = {
```
ankrgyl, line 56: can you link to the openrouter docs which use a similar mechanism?

packages/proxy/package.json
ibolmo: thanks! will make sure we do

packages/proxy/schema/index.ts
ibolmo: here's where they mention them heuristic: https://openrouter.ai/docs/use-cases/reasoning-tokens#reasoning-effort-level

google has an interesting approach: 
image

packages/proxy/package.json
```diff
@@ -82,7 +82,7 @@
     "@anthropic-ai/sdk": "^0.39.0",
     "@apidevtools/json-schema-ref-parser": "^11.9.1",
     "@aws-sdk/client-bedrock-runtime": "^3.738.0",
-    "@braintrust/core": "^0.0.85",
+    "@braintrust/core": "link:../../../sdk/core/js",
```
ibolmo, line 85: will revert this before we merge

packages/proxy/schema/index.ts
```diff
@@ -49,10 +49,79 @@ export const modelParamToModelParam: {
...
+  low: 0.2,
+  medium: 0.5,
+  high: 0.8,
+} as const;
+
+const getBudgetMultiplier = (effort: keyof typeof effortToBudgetMultiplier) => {
```
choochootrain, line 62: tests 🙏🏾

choochootrain (commented):
seems legit

packages/proxy/schema/models.ts
```diff
@@ -44,7 +44,8 @@ export const ModelSchema = z.object({
   input_cost_per_mil_tokens: z.number().nullish(),
   output_cost_per_mil_tokens: z.number().nullish(),
   displayName: z.string().nullish(),
-  o1_like: z.boolean().nullish(),
+  o1_like: z.boolean().nullish().describe('DEPRECATED use "reasoning" instead'),
```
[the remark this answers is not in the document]
ibolmo: yeah originally that was the idea but then: https://github.com/braintrustdata/braintrust/pull/5462#discussion_r2070786685

packages/proxy/types/openai.ts
```diff
@@ -0,0 +1,59 @@
+// TODO: move from core
```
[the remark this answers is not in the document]
ibolmo: I took a look and most of the usage is related to the proxy. The few other mentions all depend on the proxy already so it would be an improvement to move this closer to the owner of those types. Especially as we diverge from OpenAI and we move to our braintrust unified interface we'll want that closer to where it is needed most. Core value being that we avoid the sdk repo coupling making us more easily iterate on the interface

packages/proxy/schema/index.ts
ibolmo: huge thanks for calling this one out. would have been 🔥 

tl;dr

the order that we call the mappers mattered!

.envrc
```diff
@@ -0,0 +1,2 @@
+source_up_if_exists
+dotenv_if_exists
```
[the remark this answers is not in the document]
choochootrain: bump

packages/proxy/schema/index.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      const result = translateParams(
+        provider as ModelFormat,
+        openai as unknown as Record<string, unknown>,
+      );
+      try {
+        expect(result).toEqual(expected);
+      } catch (error) {
+        try {
+          // try to relax the output a little
+          expect(result).toMatchObject(expected);
```
choochootrain, line 142: if you want to save some typing we can use `.toMatchInlineSnapshot` and sanity check each output before merging this PR. then subsequent changes can easily be seen in the snapshots

packages/proxy/schema/index.ts
```diff
@@ -431,22 +526,31 @@ export function translateParams(
...
+    const hasDefaultParam =
       translatedKey !== undefined &&
-      defaultModelParamSettings[toProvider][translatedKey] !== undefined
-    ) {
-      translatedParams[translatedKey] = safeValue;
-    } else {
-      translatedParams[k] = safeValue;
+      defaultModelParamSettings[toProvider][translatedKey] !== undefined;
+
+    translatedParams[hasDefaultParam ? translatedKey : k] = safeValue;
```
choochootrain, line 546: when would we want to set `translatedParams[k]` to anything if `translatedKey` is undefined? this is also the case prior to your change so im curious if you found the use case. ill revisit in #217 so no need to do anything

packages/proxy/schema/index.ts
```diff
@@ -49,10 +49,91 @@ export const modelParamToModelParam: {
...
+} = {
+  anthropic: {
+    reasoning_effort: ({
+      reasoning_effort,
+      max_tokens,
+      max_completion_tokens,
+      temperature,
+      ...params
+    }) => {
+      debugger;
```
choochootrain, line 79: rm

packages/proxy/schema/models.ts
```diff
@@ -2548,7 +2553,17 @@ export const AvailableModels: { [name: string]: ModelSpec } = {
     displayName: "Command Light",
   },
 
+  // TODO: add anthropic 3.7 converse
```
choochootrain, line 2556: in this PR?

packages/proxy/types/openai.ts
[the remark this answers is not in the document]
choochootrain: makes sense, looks like there are some other things we would have to move to fully decouple. is that happening in this PR or another one?

packages/proxy/src/providers/openai.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      messages: [
+        { role: "system", content: "You are a helpful assistant." },
+        { role: "user", content: "Tell me a short joke about programming." },
+      ],
+      stream: true,
+      max_tokens: 150,
+    },
+  });
+
+  expect(json()).toEqual({
```
choochootrain, line 26: this is another good usecase for snapshots but not blocking

packages/proxy/src/proxy.ts
```diff
@@ -1323,6 +1345,7 @@ function chatCompletionMessageFromResponseOutput(
   };
 }
 
+// TODO(ibolmo): should return the reasoning
```
choochootrain, line 1348: in this PR?

packages/proxy/src/providers/anthropic.ts
```diff
@@ -350,9 +392,12 @@ export function anthropicCompletionToOpenAICompletion(
   completion: AnthropicCompletion,
   isFunction: boolean,
   isStructuredOutput: boolean,
-): ChatCompletion {
+): OpenAIChatCompletion {
+  // TODO: will we ever have text -> thinking -> text -> tool_use, thus are we dropping tokens?
```
choochootrain, line 396: good point, i would think that's the direction things are heading with the emphasis on agents. unclear if this already happens though

packages/proxy/src/providers/anthropic.test.ts
```diff
@@ -0,0 +1,201 @@
+import { describe, it, expect } from "vitest";
+import { callProxyV1 } from "../../utils/tests";
+import {
+  OpenAIChatCompletion,
+  OpenAIChatCompletionChunk,
+  OpenAIChatCompletionCreateParams,
+} from "@types";
+
+it("should convert OpenAI streaming request to Anthropic and back", async () => {
```
choochootrain, line 9: nit: this does test end to end requests but we're not really testing the request conversion itself. for example we could be incorrectly mapping the parameters or message roles and this test would still pass. maybe a different name for these tests?

also it's probably more useful to test `anthropicEventToOpenAIEvent` and `anthropicCompletionToOpenAICompletion` with snapshotted data so we can run the bulk of our unit tests without actually hitting providers

choochootrain: i think this applies for all providers - testing the side-effect free "convert from openai to X" and "convert X to openai" functions is really what we want to focus efforts on

choochootrain (commented):
this is getting there! i left a few testing notes and a couple questions about which TODOs are in scope for this PR or followups

packages/proxy/types/openai.ts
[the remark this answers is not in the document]
ibolmo: going to have to be another one. based on another internal thread we're going to want to possibly remove @braintrust/core all together. that would take care of this then

packages/proxy/schema/index.test.ts
ibolmo: Just tried this, and noticed we lost some type protections.

packages/proxy/types/openai.ts
```diff
@@ -0,0 +1,59 @@
...
+import { chatCompletionMessageParamSchema } from "@braintrust/core/typespecs/dist";
```
ankrgyl, line 2: don't leave imports in with `dist`

packages/proxy/schema/index.ts
ibolmo: yeah I haven't checked. I honestly think we need to remove this section and instead just do the translate params mappers that I started. the idea is that you get openai as input and get anthropic params as the output.

This way we centralize that kind of translation to a spot vs. right now it's a pseudo translation and then have each provider decide how to grab the data from openai-like params.

packages/proxy/schema/models.ts
ibolmo: going to have do it as a follow-up

packages/proxy/types/openai.ts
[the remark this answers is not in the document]
ankrgyl: that's not really true. we use the prompt and message schemas to manage the `Prompt` type in the SDK and in our UI. I think moving it to the proxy will require making the SDK start to depend on the proxy, which has the inverted set of challenges (we currently can conveniently use a workspace import while iterating on those).

I'd really prefer not to bite off more than we can chew in these changes, and stick to the established convention we have of keeping this in `@braintrust/core` until we explicitly decide to change things.

packages/proxy/src/providers/openai.test.ts
ibolmo: how do snapshots handle very variable data? i.e. on each request the content of the chat completion may differ or the system_fingerprint may be different

packages/proxy/src/proxy.ts
ibolmo: old comment. removed

packages/proxy/src/providers/anthropic.ts
ibolmo: yeah mostly surprised by the choice to only grab the first. need to keep an eye on user complaints on this

packages/proxy/src/providers/google.ts
```diff
@@ -253,9 +284,9 @@ export function googleCompletionToOpenAICompletion(
     object: "chat.completion",
     usage: data.usageMetadata
       ? {
-          prompt_tokens: data.usageMetadata.promptTokenCount,
-          completion_tokens: data.usageMetadata.candidatesTokenCount,
-          total_tokens: data.usageMetadata.totalTokenCount,
+          prompt_tokens: data?.usageMetadata?.promptTokenCount || 0,
```
ankrgyl, line 287: since you're checking `data.usageMetadata` above, why would this change be needed?

packages/proxy/src/providers/anthropic.test.ts
ibolmo: > bulk of our unit tests without actually hitting providers

yeah I was in favor of this as well, but @clutchski made a good point recently that we should have our tests complain if the ai providers change something under us. the tests are slower, but hopefully doing a live request may give us real actionable feedback

ibolmo: > nit: this does test end to end requests but we're not really testing the request conversion itself.

yeah this is more of a compromise because it'd be a larger scope to refactor our providers. it's a bit of a spaghetti code at the moment, and for us to really test:

openai -> anthropic (check params against actual call to anthropic)

and

anthropic -> openai (check response is valid openai)

we need to refactor heavily. i.e. it'd be gnarly. i.e. looking at anthropic we have:

```
                     vertex
                    /
openai -> anthropic - bedrock
                    \ 
                      anthropic
```

so technically there's multiple intermediate params/inputs that we need to check, but some of them like bedrock are not their specific implementation.

packages/proxy/src/providers/google.ts
ibolmo: if I recall it was a typing error that led me here

clutchski (approved):

PR #225, opened 2025-05-16
Properly base64 encode PDF attachments for openai

choochootrain:
also add descriptions for custom models


packages/proxy/src/util.ts
```diff
@@ -80,82 +81,63 @@ export class ProxyBadRequestError extends Error {
...
-    // Check if filename exists and remove fragment identifier if present
-    if (filename) {
-      try {
-        return decodeURIComponent(filename.split("#")[0]);
-      } catch (e) {
-        return filename.split("#")[0];
-      }
+    // If the URL is not http(s), file cannot be accessed
+    // If pathname is empty or ends with "/", there's no filename to extract
+    if (parsedUrl.protocol !== "http:" && parsedUrl.protocol !== "https:") {
```
ibolmo, line 103: I was under the impression that we wanted `s3://` and maybe `gs://` to work?

packages/proxy/src/util.test.ts
```diff
@@ -1,167 +1,206 @@
...
-    ).toBe("file.pdf");
+      parseFileMetadataFromUrl("http://foo.com/bar/?file=example.pdf"),
+    ).toBeUndefined();
+    expect(parseFileMetadataFromUrl("gs://bucket/file.pdf")).toBeUndefined();
+    expect(
+      parseFileMetadataFromUrl("ftp://files.org/documents/sample.pdf"),
+    ).toBeUndefined();
     expect(
-      parseFilenameFromUrl(
+      parseFileMetadataFromUrl("s3://my-bucket/backup/archive.pdf"),
```
ibolmo, line 116: is this backward incompatible?

packages/proxy/src/util.ts
choochootrain: nope, that never worked anyways since we use `fetch` to get the file contents for b64-ing

packages/proxy/src/util.test.ts
choochootrain: see above

ibolmo (approved):

PR #226, opened 2025-05-17
Fix token accounting for anthropic models

ankrgyl:
* Fix token accounting/conversion for Anthropic models' cached tokens
* Support cache breakpoints while converting openai messages to anthropic
* Fix a small bug with gemini translation


packages/proxy/src/proxy.ts
```diff
@@ -636,14 +637,27 @@ export async function proxyV1({
...
-                      completion_tokens: result.usage.completion_tokens,
+                      tokens: extendedUsage.data.total_tokens,
+                      prompt_tokens: extendedUsage.data.prompt_tokens,
+                      completion_tokens: extendedUsage.data.completion_tokens,
+                      prompt_cached_tokens:
+                        extendedUsage.data.prompt_tokens_details?.cached_tokens,
+                      prompt_cache_creation_tokens:
+                        extendedUsage.data.prompt_tokens_details
+                          ?.cache_creation_tokens,
+                      completion_reasoning_tokens:
```
ibolmo, line 658: will need to add this to the STANDARD_METRICS: https://github.com/braintrustdata/braintrust/pull/5540/files. we should see if the  detection is working, and would have caught this

packages/proxy/types/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./openai";
```
ibolmo, line 1: nice ty

packages/proxy/src/providers/anthropic.ts
```diff
@@ -176,19 +179,43 @@ export interface AnthropicCompletion {
...
 }
 
 function updateUsage(
   anthropic: z.infer<typeof anthropicUsage>,
   openai: Partial<CompletionUsage>,
 ) {
   if (!isEmpty(anthropic.input_tokens)) {
-    openai.prompt_tokens = anthropic.input_tokens;
+    // OpenAI's convention is to accumulate all input tokens, including cached tokens.
+    openai.prompt_tokens =
```
ibolmo, line 196: #nit would be nice to convert this to a pure function

ibolmo (approved):
LGTM, a small heads up

packages/proxy/src/providers/anthropic.ts
ankrgyl: hm it's intentionally not -- the point is to accumulate metrics across streamed events

PR #229, opened 2025-05-19
try a python package helper

ibolmo:
Now that we're drifting away from OpenAI's specs the python library use + our AI proxy causes unfortunate type errors if users provide or accesses data that is not part of the spec:

Screen Shot 2025-05-16 at 5 11 43 PM

This is an attempt (see [README](https://github.com/braintrustdata/braintrust-proxy/pull/229/files#diff-a84243340907809a1b57aedbd65a229d2976ea4fb5a6beb3e652cf63b38d971eR49)) to minimally affect our customer's code. In the future, we may want to consider a simpler wrap interface (like we do in SDK)


ankrgyl:
Seems like a good idea to try. But why not nest this in the existing braintrust SDK?

py/README.md
```diff
@@ -0,0 +1,130 @@
...
+response = client.chat.completions.create(
+    model="claude-3-7-sonnet-latest",
+    reasoning_effort=None,
+    stream=False,
+    messages=[
+        {
+            "role": "user",
+            "content": "How many rs in 'ferrocarril'",
+        },
+        as_openai_chat_message_param(
```
clutchski, line 49: I don't immediately understand the benefit of this. Could you be more specific in the docs?

py/README.md
```diff
@@ -0,0 +1,130 @@
...
+        ),
+        {
+            "role": "user",
+            "content": "How many e in what you said?",
+        },
+    ],
+)
+
+for event in stream:
+    delta = from_openai_chat_completion_choice_delta(event.choices[0].delta)
```
clutchski, line 124: same can you explain this?

PR #245, opened 2025-06-08
Fix parallel tool call parsing in the proxy

ankrgyl:


packages/proxy/src/proxy.ts
```diff
@@ -714,19 +714,26 @@ export async function proxyV1({
...
                         type: delta.tool_calls[0].type,
                         function: delta.tool_calls[0].function,
                       },
-                    ];
-                  } else if (tool_calls[0].function) {
-                    // TODO: what about parallel calls?
-                    tool_calls[0].function.arguments =
-                      (tool_calls[0].function.arguments ?? "") +
+                    ]);
+                  } else if (lastTool.function) {
```
ibolmo, line 734: may want to see if this works:

```ts
if (delta.tool_calls?.length) {
  const lastTool = tool_calls?.[tool_calls?.length - 1];

  // do we need to add the tool call to our list?
  if (lastTool?.id !== delta.tool_calls?.[0]?.id) {
    tool_calls = (tool_calls ?? []).concat([
      {
        index: 0, // should this be tool_calls.length  ?
        id: delta.tool_calls[0].id,
        type: delta.tool_calls[0].type,
        function: delta.tool_calls[0].function,
      },
    ]);
    // same tool or no tool call
  } else if (lastTool?.function) {
    lastTool.function.arguments =
      (lastTool.function.arguments ?? "") +
      (delta.tool_calls[0].function?.arguments ?? "");
  }
}
```

ibolmo (approved):

packages/proxy/src/proxy.ts
ibolmo: would be good to see a few tests on this (esp. locally)

ankrgyl: cleaned this up

PR #247, opened 2025-06-08
Capture token limits for each model

ankrgyl:


ibolmo (approved):
nice. I take it this will eventually show up as a new stat in the product for the model card

PR #248, opened 2025-06-10
Tweak provider error message

edenh:


packages/proxy/src/proxy.ts
```diff
@@ -1139,7 +1139,7 @@ async function fetchModelLoop(
         headersString.push(`${key}: ${value}`);
       });
       const errorText =
-        `AI provider returned ${httpCode} error.\n\nHeaders:\n` +
+        `The AI provider returned ${httpCode} error. This is likely not an issue with Braintrust, but rather with the model provider.${httpCode === 500 ? " Please check the model provider's status page." : httpCode === 429 ? " Please adjust max concurrency." : ""}\n\nHeaders:\n` +
```
ibolmo, line 1142: The "please adjust concurrency" is throwing me off. Is that on the Braintrust side (i.e. playground concurrency) or is that like a quota/limit problem in the AI provider? 

You may consider just dropping this one and keep the rest. I think most folks know what 429 is.

Would you see us return the statusText and/or `response.text()` in addition to headers?

ibolmo: I'm also curious if we shouldn't return a JSON message rather than a plain message. This way we avoid losing fidelity.

PR #250, opened 2025-06-11
fix anthropic and bedrock system only calls

ibolmo:


packages/proxy/src/providers/anthropic.test.ts
```diff
@@ -1,11 +1,48 @@
-import { describe, it, expect } from "vitest";
+import { it, expect } from "vitest";
 import { callProxyV1 } from "../../utils/tests";
 import {
   OpenAIChatCompletion,
   OpenAIChatCompletionChunk,
   OpenAIChatCompletionCreateParams,
 } from "@types";
 
+it.only("should convert single system message OpenAI request to Anthropic and back", async () => {
```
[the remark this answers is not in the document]
colsondonohue: diamond's got a point here

packages/proxy/src/proxy.ts
```diff
@@ -2110,12 +2110,17 @@ async function fetchAnthropicChatCompletions({
...
-  for (const m of oaiMessages as Message[]) {
+  for (const message of oaiMessages as Message[]) {
+    let m = message;
     let role: MessageRole = m.role;
     let content: any = await openAIContentToAnthropicContent(m.content);
     if (m.role === "system") {
       system = content;
-      continue;
+
+      // hack: anthropic requires at least one user message. could do something smarter, but shouldn't have an affect
```
colsondonohue: nit: 
```suggestion
      // hack: anthropic requires at least one user message. could do something smarter, but shouldn't have an effect
```

packages/proxy/src/proxy.ts
```diff
@@ -2110,12 +2110,17 @@ async function fetchAnthropicChatCompletions({
...
+      // @ts-expect-error
+      role = m.role = "user";
```
colsondonohue, line 2122: could maybe do `"user" as const` here and not need the `@ts-expect-error`

ibolmo: yeah I think it's because it knows it should be a system message 😂  

image

PR #252, opened 2025-06-12
Nicer errors for persistent 400-class errors

ankrgyl:
We were previously propagating these error responses directly, but now clarify that they are from the AI provider itself.

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -1061,8 +1061,9 @@ async function fetchModelLoop(
         cachePut,
       );
       secretName = secret.name;
-      // If the response is ok, we can break out of the loop.
-      if (proxyResponse.response.ok) {
+      // If the response is ok or a 400 (Bad Request), we can break out of the loop and return
+      // the exact response.
+      if (proxyResponse.response.ok || proxyResponse.response.status === 400) {
         errorHttpCode = undefined;
         errorHttpHeaders = new Headers();
         break;
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -1073,7 +1073,7 @@ async function fetchModelLoop(
           !TRY_ANOTHER_ENDPOINT_ERROR_CODES.includes(
             proxyResponse.response.status,
           )) ||
-        // Or we've exhausted the set of secrets
+        // Or we have not exhausted the set of secrets
         i < secrets.length - 1
         // Then mark this as an error response, and fall through to the error handling logic.
       ) {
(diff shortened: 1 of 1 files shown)


packages/proxy/src/proxy.ts
```diff
@@ -1061,32 +1061,38 @@ async function fetchModelLoop(
...
       );
       secretName = secret.name;
-      if (
-        proxyResponse.response.ok ||
+      // If the response is ok, we can break out of the loop.
+      if (proxyResponse.response.ok) {
+        errorHttpCode = undefined;
+        errorHttpHeaders = new Headers();
+        break;
+      } else if (
```
ibolmo, line 1070: (nit) you can avoid the `else` since we broke earlier

packages/proxy/src/proxy.ts
```diff
@@ -1061,32 +1061,38 @@ async function fetchModelLoop(
...
+        // Otherwise, if this is a non-retryable 400-error
         (proxyResponse.response.status >= 400 &&
           proxyResponse.response.status < 500 &&
           !TRY_ANOTHER_ENDPOINT_ERROR_CODES.includes(
```
ibolmo, line 1074: (future) we should probably delegate this determination to the providers that we have defined. i.e. I found that the codes vary by provider in my past life.

packages/proxy/src/proxy.ts
```diff
@@ -1126,49 +1132,64 @@ async function fetchModelLoop(
...
+      const existingResponse = proxyResponse;
       proxyResponse = {
-        response: new Response(null, { status: httpCode }),
+        response: new Response(null, { status: errorHttpCode }),
         stream: new ReadableStream({
           start(controller) {
             controller.enqueue(new TextEncoder().encode(errorText));
-            controller.close();
+
+            if (existingResponse?.stream) {
```
ibolmo, line 1158: oh nice catch. i wonder how it was working before

packages/proxy/src/proxy.ts
```diff
@@ -1061,32 +1061,38 @@ async function fetchModelLoop(
...
+      } else if (
+        // Otherwise, if this is a non-retryable 400-error
         (proxyResponse.response.status >= 400 &&
           proxyResponse.response.status < 500 &&
           !TRY_ANOTHER_ENDPOINT_ERROR_CODES.includes(
             proxyResponse.response.status,
-          ))
+          )) ||
+        // Or we've exhausted the set of secrets
+        i < secrets.length - 1
```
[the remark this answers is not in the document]
ibolmo: :nod:

packages/proxy/src/proxy.ts
ankrgyl: oh no we need it because if it's a retryable error, we dont want to return an error yet.

packages/proxy/src/proxy.ts
ankrgyl: for sure

PR #259, opened 2025-06-20
Revert "fix anthropic and bedrock system only calls (#250)"

ankrgyl:
This reverts commit 809e00b6e28784c54ef78c92aeebc5adc3468af6.


packages/proxy/schema/index.ts
```diff
@@ -489,7 +489,7 @@ export function translateParams(
 
     const hasDefaultParam =
       translatedKey !== undefined &&
-      defaultModelParamSettings[toProvider]?.[translatedKey] !== undefined;
+      defaultModelParamSettings[toProvider][translatedKey] !== undefined;
```
ibolmo: May want to keep this or as a follow-up

ibolmo (approved):

packages/proxy/schema/index.ts
ankrgyl: will do

PR #260, opened 2025-06-20
Show headers below response error text

ankrgyl:


packages/proxy/src/proxy.ts
```diff
@@ -2281,6 +2282,9 @@ async function fetchAnthropicChatCompletions({
     delete params.model;
   }
 
+  console.log("SYSTEM", system);
+  console.log("MESSAGES", messages);
```
choochootrain: are these to keep?

ankrgyl: oops

packages/proxy/src/proxy.ts
```diff
@@ -1145,9 +1145,8 @@ async function fetchModelLoop(
       errorHttpHeaders.forEach((value, key) => {
         headersString.push(`${key}: ${value}`);
       });
-      const errorText =
-        `AI provider returned ${errorHttpCode} error.\n\nHeaders:\n` +
-        headersString.join("\n");
+      const errorText = `AI provider returned ${errorHttpCode} error:`;
+      const headersSuffix = `\n\nHeaders:\n` + headersString.join("\n");
```
ibolmo, line 1149: starting to think we should send it in a fell formed structure so that the UI is responsible to show the best representation. we had a similar problem at zapier, and ultimately we wanted to avoid being in the middle as much as possible. 

i.e. it would be ideal if we could give the user the literal request so that they can curl/try it themselves to prove it wasn't us

ankrgyl: Yep I think we should keep evolving it towards this

PR #272, opened 2025-07-29
add Gemini 2.5 Flash-Lite and bump models

choochootrain:
also bump `claude-2.1` to `claude-3-haiku-20240307` for end to end tests


ibolmo (approved):
untested, but LGTM

PR #279, opened 2025-08-08
Dynamically fetch models

ankrgyl:


choochootrain (approved):
seems fine, we may want to make it clear that making breaking changes to `ModelSpec` requires you to deploy the proxy shortly after landing to avoid this refresh thing from breaking

packages/proxy/schema/models.ts
```diff
@@ -84,5 +85,68 @@ export const ModelSchema = z.object({
 
 export type ModelSpec = z.infer<typeof ModelSchema>;
 
-import models from "./model_list.json";
-export const AvailableModels = models as { [name: string]: ModelSpec };
+import modelListJson from "./model_list.json";
+const modelListJsonTyped = modelListJson as { [name: string]: ModelSpec };
```
manugoyal: Can we use zod here?

packages/proxy/schema/models.ts
```diff
@@ -84,5 +85,68 @@ export const ModelSchema = z.object({
...
+
+declare global {
+  var availableModels: { [name: string]: ModelSpec } | undefined;
```
manugoyal: I don't understand why we need `declare global` here. The internet suggests that this construct is used to inform typescript about global variables that may have been created / modified in other modules. But it looks like we're only using `availableModels` in this file, so can we confine it to this file, similar to `cachedModels` and friends?

If there are complicated JS bundling issues that require this, then an explanatory comment would be helpful.

packages/proxy/schema/models.ts
```diff
@@ -84,5 +85,68 @@ export const ModelSchema = z.object({
...
+
+  const fetchUrl = _urljoin(appUrl, "api/models/model_list.json");
+
+  try {
+    const response = await fetch(fetchUrl);
+    if (!response.ok) {
+      throw new Error(`Failed to fetch models: ${response.statusText}`);
+    }
+    const data = await response.json();
+    cachedModels = data as { [name: string]: ModelSpec };
```
manugoyal: zod here too

packages/proxy/schema/models.ts
```diff
@@ -84,5 +85,68 @@ export const ModelSchema = z.object({
...
+  if (isCacheValid()) {
+    return;
+  }
+
+  const dynamicModels = await loadModelsFromControlPlane(appUrl);
+  if (dynamicModels) {
+    if (!globalThis.availableModels) {
+      globalThis.availableModels = { ...modelListJsonTyped };
+    }
+    Object.assign(globalThis.availableModels, dynamicModels);
```
manugoyal: Curious why not just overwrite `globalThis.availableModels`? Do we intentionally not want model removals to be live-propagated? Don't really have a strong opinion, but just curious.

packages/proxy/schema/models.ts
```diff
@@ -84,5 +85,68 @@ export const ModelSchema = z.object({
...
+  } catch (error) {
+    console.warn(
+      `Failed to load models dynamically from control plane (${fetchUrl}), falling back to static import:`,
+      error,
+    );
+  }
+  return cachedModels;
+}
+
+// Initialize models on startup and refresh when expired
```
manugoyal: I don't really understand how this works. Is the expectation that in any code where the users want dynamically-refreshed models, they must ensure `initializeModels` is called beforehand, and frequently enough to keep this refreshed? Currently I only see this function called in one place in the codebase, and it's unclear to me if it's actually called before every usage of `getAvailableModels`.

If so, it feels hard to be confident that we're really refreshing the models before any time we call `getAvailableModels`. Would it make sense to do the slightly-cumbersome thing of having `getAvailableModels` be an async function that refreshes the cached models whenever it needs to? Or alternatively, could we put this in every server entrypoint and have it refresh in a background loop or something?

PR #283, opened 2025-08-13
Custom bedrock endpoint

ankrgyl:


packages/proxy/src/providers/bedrock.ts
```diff
@@ -517,7 +532,14 @@ export async function fetchConverse({
   let system: SystemContentBlock[] | undefined = undefined;
   for (const m of oaiMessages as OaiMessage[]) {
     if (m.role === "system") {
-      system = [{ text: m.content }];
+      system = [
+        {
+          text:
+            typeof m.content === "string"
+              ? m.content
+              : JSON.stringify(m.content),
```
ibolmo, line 540: did you run into something when testing?

ankrgyl: just a lingering type error

PR #287, opened 2025-08-21
Add helper to override parameters based on key & model

knjiang:


packages/proxy/schema/index.ts
```diff
@@ -80,6 +80,38 @@ export const sliderSpecs: {
   topK: [1, 100, 1, false],
 };
 
+// Format-specific slider specification overrides
```
choochootrain, line 83: hm should we override the other way so only providers with >1 temperature support are in here?

choochootrain: ah i see, multiple providers support >1 temperature, disregard

choochootrain (approved):

PR #292, opened 2025-09-08
Responses API improvements

ankrgyl:


packages/proxy/src/proxy.ts
```diff
@@ -815,6 +816,22 @@ export async function proxyV1({
               }
               break;
             }
+            case "response": {
```
ankrgyl, line 834: @CLowbrow FYI

packages/proxy/src/proxy.ts
```diff
@@ -2979,5 +3009,13 @@ function logSpanInputs(
       });
       break;
     }
+    case "response": {
```
ankrgyl, line 3027: @CLowbrow FYI

PR #293, opened 2025-09-09
Propagate abort signal through proxy

ankrgyl:


packages/proxy/src/providers/bedrock.ts
```diff
@@ -498,9 +498,11 @@ function translateTools(tools: ChatCompletionTool[]): ToolConfiguration {
 export async function fetchConverse({
   secret,
   body,
+  signal,
 }: {
   secret: APISecret;
   body: Record<string, unknown>;
+  signal?: AbortSignal;
```
choochootrain, line 505: im guessing the aws sdk doesnt let you thread this through?

ankrgyl: it actually does. good catch

PR #298, opened 2025-09-15
Add reasoning params

CLowbrow:
These weren't getting passed through to try prompt before.

### DIFF
--- a/packages/proxy/schema/index.ts
+++ b/packages/proxy/schema/index.ts
@@ -54,6 +54,8 @@ export const modelParamToModelParam: {
   parallel_tool_calls: null,
   response_format: null,
   reasoning_effort: "reasoning_effort",
+  reasoning_enabled: "reasoning_enabled",
+  reasoning_budget: "reasoning_budget",
   verbosity: "verbosity",
   stop: null,
 };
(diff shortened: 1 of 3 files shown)


packages/proxy/schema/index.ts
```diff
@@ -54,6 +54,8 @@ export const modelParamToModelParam: {
   parallel_tool_calls: null,
   response_format: null,
   reasoning_effort: "reasoning_effort",
+  reasoning_enabled: "reasoning_enabled",
```
ibolmo, line 57: i think these should be `null`. I'm trying to connect the dots but when these are not null we will attempt to include them automatically when we proxy to other providers. did you happen to make the change and then try to run a call to say gemini or openai (non anthropic)

ibolmo: nevermind I think we're good. looks like [this](https://github.com/braintrustdata/braintrust-proxy/blob/24b617109becad4c88fb55ae7e4625d01193ff13/packages/proxy/schema/index.ts#L148-L149) will avoid any flavors without the params from being included, i believe.

ibolmo (approved):
you may want to double check calling from the playground (make sure this is your local and not on a preview) against other providers with and without the params

PR #302, opened 2025-09-28
avoid manual json schema to gemini (openapi 3) schema

ibolmo:
fixes long standing $schema and other schema related issues like raising an exception for unions

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -1,5 +1,4 @@
 import { MessageParam } from "@anthropic-ai/sdk/resources";
-import $RefParser from "@apidevtools/json-schema-ref-parser";
 import {
   type ChatCompletionMessageParamType as Message,
   type MessageRoleType as MessageRole,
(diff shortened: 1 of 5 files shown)


packages/proxy/utils/tests.ts
```diff
@@ -35,18 +35,17 @@ export function createResponseStream(): [
...
-  const setStatusCode = (code: number) => {
-    statusCode = code;
+  const ref = {
+    headers: {} as Record<string, string>,
+    statusCode: -1,
+    setHeader(name: string, value: string) {
+      ref.headers[name] = value;
+    },
+    setStatusCode(code: number) {
+      ref.statusCode = code;
```
ibolmo, line 45: I found this wasn't getting updated. Using an object in JS and passing it around is as good as passing with reference.

packages/proxy/src/proxy.ts
```diff
@@ -569,8 +569,9 @@ export async function proxyV1({
     );
     stream = proxyStream;
 
+    console.log("proxyResponse", proxyResponse.status);
```
ibolmo, line 572:
```suggestion
```

PR #307, opened 2025-09-30
add google (genai) provider and types

ibolmo:
Part of a bigger ticket to add genai support to our product.


packages/proxy/schema/index.test.ts
```diff
@@ -1,5 +1,5 @@
 import { MessageCreateParamsBase } from "@anthropic-ai/sdk/resources/messages";
-import { GenerateContentParameters } from "@google/genai";
+import { GenerateContentParameters } from "../types/google";
```
ibolmo, line 2: we no longer need a direct dependency now that types/google.ts exists. also avoids type mismatch errors.

packages/proxy/src/providers/google.ts
```diff
@@ -191,51 +196,59 @@ export function googleEventToOpenAIChatEvent(
...
-                })) || [];
-            return {
-              index: 0,
-              delta: {
-                role: "assistant",
-                content: firstText?.text ?? "",
-                tool_calls: toolCalls.length > 0 ? toolCalls : undefined,
-                ...(firstThought && {
-                  reasoning: {
+      ? (() => {
```
ibolmo, line 201: I think the llm may have done too much here. I'll add more tests.

packages/proxy/types/google.ts
```diff
@@ -0,0 +1,642 @@
+import { z } from "zod";
+
+export const finishReasonSchema = z.enum([
```
ibolmo, line 4: mostly generated by the llm, and i manually reviewed. i think i had to change 20 lines so pretty quick.

CLowbrow (approved):

PR #311, opened 2025-10-02
add gemini-2.5-flash and lite 09-2025

ibolmo:


packages/proxy/schema/model_list.json
```diff
@@ -2211,6 +2211,20 @@
     "max_input_tokens": 1048576,
     "max_output_tokens": 65535
   },
+  "gemini-2.5-flash-preview-09-2025": {
```
ibolmo, line 2214: qa pending

ibolmo: QA looks good.

PR #314, opened 2025-10-06
add codex and make sure to use responses for gpt-5-pro and codex

ibolmo:


packages/proxy/schema/model_list.json
```diff
@@ -71,6 +49,40 @@
     "max_input_tokens": 400000,
     "max_output_tokens": 128000
   },
+  "gpt-5-pro": {
```
ibolmo, line 52: demoting from the first pick since we are doing fake responses and our streaming support is lack luster

edenh (approved):

PR #431, opened 2026-04-01
Added glm-4p7 and qwen3-vl-30b-a3b-instruct

josep-foronda:
Got the details from:
https://fireworks.ai/models/fireworks/glm-4p7
https://fireworks.ai/models/fireworks/qwen3-vl-30b-a3b-instruct
https://docs.z.ai/guides/llm/glm-4.7


aswink (approved):
Worth double checking that running sync-models won't overwrite this change.

Unfortunately seems like this model isn't in https://raw.githubusercontent.com/BerriAI/litellm/refs/heads/main/litellm/model_prices_and_context_window_backup.json yet, not sure how we typically handle missing values

PR #435, opened 2026-04-08
model_list.json: add supported_regions to vertex models

aswink:
Do this by scraping the gcp vertex docs, which seems as good a way to do this
as any. Also ran the script to update the model list.

PR #436, opened 2026-04-08
sync-models: add --major-only heuristic

aswink:
Add a heuristic to add new major models from openai/anthropic/google to
model_list.json, assuming that litellm's list already contains those models.
Explicitly excludes a bunch of old models from these providers that we haven't
added yet, to avoid polluting the model list. I ran this and updated the model
list too.

### DIFF
--- a/packages/proxy/schema/index.ts
+++ b/packages/proxy/schema/index.ts
@@ -447,6 +447,19 @@ export const AvailableEndpointTypes: { [name: string]: ModelEndpointType[] } = {
   "grok-2-1212": ["xAI"],
   "grok-vision-beta": ["xAI"],
   "grok-beta": ["xAI"],
+  "gemini-flash-latest": ["google"],
+  "gemini-flash-lite-latest": ["google"],
+  "gemini-pro-latest": ["google"],
+  "gpt-5.2-pro-2025-12-11": ["openai", "azure"],
+  "gpt-5.4-pro-2026-03-05": ["openai", "azure"],
+  "vertex_ai/gemini-3-flash-preview": ["vertex"],
+  "vertex_ai/gemini-3-pro-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-pro-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-pro-preview-customtools": ["vertex"],
+  "vertex_ai/gemini-2.5-flash-image": ["vertex"],
+  "vertex_ai/gemini-3-pro-image-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-flash-image-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-flash-lite-preview": ["vertex"],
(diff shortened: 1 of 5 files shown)


aswink:
* **#436** <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/436?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#435** <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

cpinn:
Is there a reason we don't want to add all possible models? 

I ran the current sync_models script a while ago and it did seem like it would add a ton of models we don't list as supported. 

I was also wondering is there a reason we don't always want to add specific versioned models?

packages/proxy/schema/index.ts
```diff
@@ -447,6 +447,19 @@ export const AvailableEndpointTypes: { [name: string]: ModelEndpointType[] } = {
...
+  "gpt-5.2-pro-2025-12-11": ["openai", "azure"],
+  "gpt-5.4-pro-2026-03-05": ["openai", "azure"],
+  "vertex_ai/gemini-3-flash-preview": ["vertex"],
+  "vertex_ai/gemini-3-pro-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-pro-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-pro-preview-customtools": ["vertex"],
+  "vertex_ai/gemini-2.5-flash-image": ["vertex"],
+  "vertex_ai/gemini-3-pro-image-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-flash-image-preview": ["vertex"],
+  "vertex_ai/gemini-3.1-flash-lite-preview": ["vertex"],
```
knjiang, line 462: are these valid names? i thought they use publishers/models/ for vertex

aswink:
> Is there a reason we don't want to add all possible models?
> 
> I ran the current sync_models script a while ago and it did seem like it would add a ton of models we don't list as supported.
> 
> I was also wondering is there a reason we don't always want to add specific versioned models?

I tried it at first, but it was a pretty large list of models and most of them seemed old/outdated, so I didn't want to do it all as part of this? We can remove anything from the exclude list at any time, I just didn't want to do it here. As part of this change I really just wanted to exclude all of the "old" models as of now, so we only get new ones going forward. I'll update the exclude list to be more clear and just explicitly list all model names, rather than listing prefixes, suffixes, etc. My main goal was to automatically get new model releases if we run this periodically

aswink:
adding everything instead https://github.com/braintrustdata/braintrust-proxy/pull/456/changes

PR #438, opened 2026-04-10
Model agent sync

cpinn:
### Description

Eventually we would like more real time updates to the models we support both in product in the app and in product in the gateway. 

This PR introduces two new github actions jobs.

1. A temp job for now that aims to keep our model_list more in sync with model updates
2. A job that attempts to search for unsupported models

I did not add anything yet to add new models, our current sync_models script adds too many models without some filtering changes and I didn't want the searching job to also we responsible for the fixes at this time.

### DIFF
--- a/.github/workflows/agent-provider-model-sync.yaml
+++ b/.github/workflows/agent-provider-model-sync.yaml
@@ -1,0 +1,209 @@
+name: Agent provider model sync
+
+on:
+  schedule:
+    - cron: "30 7 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: agent-provider-model-sync
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+  id-token: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    env:
+      BRANCH_NAME: chore/agent-provider-model-sync
+      COMMIT_MESSAGE: "chore: experimental provider-direct model sync"
+      PR_TITLE: "chore: experimental provider-direct model sync"
+      PR_BODY: "Experimental automated sync that attempts to update the proxy model catalog from official provider sources directly."
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 20
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Setup pnpm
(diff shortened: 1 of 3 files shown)


packages/proxy/schema/index.ts
```diff
@@ -607,7 +607,9 @@ export const AvailableEndpointTypes: { [name: string]: ModelEndpointType[] } = {
   "deepseek-ai/DeepSeek-V3-0324": ["baseten"],
   "publishers/moonshotai/models/kimi-k2.5": ["azure"],
   "fireworks_ai/accounts/fireworks/models/glm-4p7": ["fireworks"],
-  "fireworks_ai/accounts/fireworks/models/qwen3-vl-30b-a3b-instruct": ["fireworks"]
+  "fireworks_ai/accounts/fireworks/models/qwen3-vl-30b-a3b-instruct": [
```
cpinn, line 610: The linter was failing so I had to change this.

.github/workflows/sync-models.yaml
```diff
@@ -0,0 +1,97 @@
+name: Update models (temporary)
```
cpinn, line 1: @aswink I was working on a github action job to run sync_models.

I ran into the same issues you were seeing and realized you were working in this area as well. I just added the update for now and wanted to know what you had been thinking about with that script/the future.

aswink: Oh this seems better than what I had, I'm happy to drop https://github.com/braintrustdata/braintrust-proxy/pull/436 and we can use this instead?

.github/workflows/agent-provider-model-sync.yaml
```diff
@@ -0,0 +1,174 @@
...
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ github.token }}
+          show_full_output: "true"
+          display_report: "true"
+          timeout_minutes: "25"
+          claude_args: |
+            --model claude-opus-4-6
+            --max-turns 200
+            --allowedTools "Read,Glob,Grep,LS,WebSearch,WebFetch,mcp__github__get_issue,mcp__github__get_issue_comments,mcp__github__search_issues,mcp__github__list_issues,mcp__github__create_issue"
```
aswink, line 39: should this just create PRs? or safer to create issues first? I don't have a strong preference either way though

aswink (approved):

.github/workflows/sync-models.yaml
cpinn: I like what you have as well because currently the github action will only update models, so gets updated prices for all existing models but from my understanding your change should allow automation of new models through add-model so I think we will need both for the future.

.github/workflows/agent-provider-model-sync.yaml
cpinn: I want to see what this prompt can find as issues first before attempting to get it to make updates. I think there will be tweaks with the prompt and the PR creation especially if the script finds new models or providers.

PR #439, opened 2026-04-10
Use locations instead of supported_regions

aswink:
I just noticed that we already had a locations field, so revert the
supported_regions change I made in favor of locations, and actually update
those vertex models to have the correct locations. Also update them to add
available_providers to ["vertex"].

### DIFF
--- a/packages/proxy/schema/model_list.json
+++ b/packages/proxy/schema/model_list.json
@@ -2924,9 +2924,6 @@
     "displayName": "Gemini 3.1 Pro (Preview)",
     "reasoning": true,
     "reasoning_budget": true,
-    "supported_regions": [
-      "global"
-    ],
     "max_input_tokens": 1048576,
     "max_output_tokens": 65536,
     "available_providers": [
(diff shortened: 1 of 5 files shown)


aswink:
* **#439** <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/proxy/schema/model_list.json
```diff
@@ -2924,9 +2924,6 @@
     "displayName": "Gemini 3.1 Pro (Preview)",
     "reasoning": true,
     "reasoning_budget": true,
-    "supported_regions": [
-      "global"
-    ],
```
knjiang, line 2929: should we specify locations for specific models?

aswink: after your comment on the other PR I realized that this wasn't vertex, this is google, which is why this was removed here

knjiang (approved):

PR #442, opened 2026-04-16
Native Inference Billing

cpinn:
Reapply native inference billing to the proxy.

Properly ensure only models with endpoint type braintrust are treated as native inference.

Refer to second commit: https://github.com/braintrustdata/braintrust-proxy/pull/442/changes/7038729a71a9ff295f5146bf4620f6a86ee4d118

### DIFF
--- a/apis/cloudflare/src/billing.ts
+++ b/apis/cloudflare/src/billing.ts
@@ -1,0 +1,86 @@
+import { type BillingEvent } from "@braintrust/proxy";
+
+const DEFAULT_BILLING_TELEMETRY_URL =
+  "https://api.braintrust.dev/billing/telemetry/ingest";
+
+function buildPayloadEvent(event: BillingEvent) {
+  if (!event.model) {
+    console.warn("billing event skipped: missing model");
+    return null;
+  }
+  if (!event.isNativeInference) {
+    return null;
+  }
+
+  if (!event.org_id) {
+    console.warn("billing event skipped: missing org_id");
+    return null;
+  }
+  if (!event.resolved_model) {
+    console.warn("billing event skipped: missing resolved_model");
+    return null;
+  }
+  const hasTokenUsageData =
+    event.input_tokens !== undefined ||
+    event.output_tokens !== undefined ||
+    event.cached_input_tokens !== undefined ||
+    event.cache_write_input_tokens !== undefined;
+  if (!hasTokenUsageData) {
+    console.warn("billing event skipped: missing token usage");
+    return null;
+  }
+  const requestId = crypto.randomUUID();
+  const timestamp = new Date().toISOString();
+
+  return {
+    event_name: "NativeInferenceTokenUsageEvent",
+    external_customer_id: event.org_id,
+    timestamp,
+    idempotency_key: requestId,
+    properties: {
+      model: event.model,
+      resolved_model: event.resolved_model,
+      org_id: event.org_id,
+      input_tokens: event.input_tokens,
+      output_tokens: event.output_tokens,
+      cached_input_tokens: event.cached_input_tokens,
+      cache_write_input_tokens: event.cache_write_input_tokens,
+    },
+  };
+}
+
+export async function sendBillingTelemetryEvent({
+  telemetryUrl,
+  event,
+}: {
+  telemetryUrl?: string;
+  event: BillingEvent;
+}): Promise<void> {
+  try {
+    const payloadEvent = buildPayloadEvent(event);
+    if (!payloadEvent) {
+      return;
+    }
+
+    const destination = telemetryUrl || DEFAULT_BILLING_TELEMETRY_URL;
+    const response = await fetch(destination, {
+      method: "POST",
+      headers: {
+        Authorization: `Bearer ${event.auth_token}`,
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({
+        events: [payloadEvent],
+      }),
+    });
+
+    if (!response.ok) {
+      const responseBody = await response.text();
+      console.warn(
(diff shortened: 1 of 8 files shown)


knjiang (approved):

apis/cloudflare/src/billing.ts
```diff
@@ -0,0 +1,86 @@
+import { type BillingEvent } from "@braintrust/proxy";
+
+const DEFAULT_BILLING_TELEMETRY_URL =
+  "https://api.braintrust.dev/billing/telemetry/ingest";
```
ibolmo, line 4: i'd remove the default/fallback. we're also disclosing this internal endpoint to a public repo. (the latter point is not a major risk/concern, though)

apis/cloudflare/src/proxy.ts
```diff
@@ -199,6 +216,17 @@ export async function handleProxyV1(
     spanLogger,
     spanId,
     spanExport,
+    billingOrgId,
+    onBillingEvent: (event) => {
+      ctx.waitUntil(
+        sendBillingTelemetryEvent({
```
ibolmo, line 222: how often is onBillingEvent called? maybe we should consider batch posting per org.

apis/cloudflare/src/billing.ts
```diff
@@ -0,0 +1,86 @@
...
+  event: BillingEvent;
+}): Promise<void> {
+  try {
+    const payloadEvent = buildPayloadEvent(event);
+    if (!payloadEvent) {
+      return;
+    }
+
+    const destination = telemetryUrl || DEFAULT_BILLING_TELEMETRY_URL;
+    const response = await fetch(destination, {
```
ibolmo, line 66: instead of a default we can return early here to disable telemetry reporting, or perhaps closer to onBillingEvent to avoid queuing up noop promises

apis/cloudflare/wrangler-template.toml
```diff
@@ -28,10 +28,12 @@ head_sampling_rate = 0.2
 # You should not need to edit this
 BRAINTRUST_APP_URL = "https://www.braintrust.dev"
 METRICS_LICENSE_KEY="<YOUR_METRICS_LICENSE_KEY>"
+BILLING_TELEMETRY_URL="https://api.braintrust.dev/billing/telemetry/ingest"
```
ibolmo, line 31: i'd move this to cloudflare. we wouldn't want all proxy customers (out in the world) posting their events.

packages/proxy/src/proxy.ts
```diff
@@ -568,6 +588,10 @@ export async function proxyV1({
   }
 
   let responseFailed = false;
+  let isNativeInference = false;
```
ibolmo, line 591: curious why we had to pull these out of the closure

packages/proxy/src/proxy.ts
```diff
@@ -759,6 +785,11 @@ export async function proxyV1({
   if (stream) {
     let first = true;
     const allChunks: Uint8Array[] = [];
+    let resolvedModel: string | undefined = undefined;
+    let inputTokens: number | undefined = undefined;
```
ibolmo, line 789: ditto here.

packages/proxy/src/proxy.ts
```diff
@@ -299,6 +318,7 @@ export async function proxyV1({
   );
 
   let orgName: string | undefined = proxyHeaders[ORG_NAME_HEADER] ?? undefined;
+  let resolvedOrgName: string | undefined = orgName;
```
ibolmo, line 321: # nit i'd move this definition closer to where it is used

packages/proxy/src/proxy.ts
```diff
@@ -649,6 +670,7 @@ export async function proxyV1({
 
         if (secrets.length > 0 && !orgName && secrets[0].org_name) {
           baseAttributes.org_name = secrets[0].org_name;
+          resolvedOrgName = secrets[0].org_name;
```
ibolmo, line 673: #nit move one line up and reuse `baseAttributes.org_name = resolvedOrgName;`

packages/proxy/src/proxy.ts
```diff
@@ -1256,6 +1342,10 @@ async function fetchModelLoop(
   }
 
   const initialIdx = getRandomInt(secrets.length);
+  const nativeInferenceSecret =
+    model === null || model === undefined
+      ? undefined
+      : secrets.find((secret) => isNativeInferenceSecret(secret, model));
```
ibolmo, line 1348: #nit looks like `isNativeInferenceSecret` already returns false for model null/undefined.

packages/proxy/src/proxy.ts
```diff
@@ -669,6 +691,10 @@ export async function proxyV1({
       signal,
       fetch,
     );
+    proxyResponse = fetchResult.modelResponse.response;
```
ibolmo, line 694: these should be a `const`

packages/proxy/src/proxy.ts
```diff
@@ -669,6 +691,10 @@ export async function proxyV1({
...
+    proxyStream = fetchResult.modelResponse.stream;
+    secretName = fetchResult.secretName;
+    isNativeInference = fetchResult.isNativeInference;
     stream = proxyStream;
```
ibolmo, line 698: seems odd to define proxyStream to just rename to stream 🤔

packages/proxy/src/proxy.ts
```diff
@@ -1089,6 +1149,28 @@ export async function proxyV1({
         });
 
         spanLogger?.end();
+        if (!responseFailed) {
```
ibolmo, line 1152: i wonder if we need to charge for the attempt anyway? we should check if openai when you make an invalid call if that still consumes some tokens.

PR #447, opened 2026-04-17
Drop temp when calling claude opus 4.7

CLowbrow:
it doesnt support it.


aswink (approved):
is there a playground change to make too?

edit: never mind, we decided to fix this in proxy/lingua

ibolmo (approved):
LGTM. I'd also do a pass/grep/ai to make sure we don't add temperature for other reasons. Maybe cache is set, or something else. Unlikely, but for sanity sake.

PR #455, opened 2026-04-21
proxy: strip "fireworks_ai/" prefix from model ids before forwarding

josep-foronda:
Catalog entries sourced from LiteLLM (via sync_models.ts) use the "fireworks_ai/accounts/fireworks/models/X" naming convention, but Fireworks' native API only accepts "accounts/fireworks/models/X". The proxy was forwarding bodyData.model verbatim, so every prefixed entry 404'd end-to-end upstream.

This mirrors the existing publishers/<x>/models/ rewrite for Vertex in the same file, scoped to the fireworks provider only.

Impact:

- Unblocks 33 fireworks_ai/* entries in model_list.json. 19 of them have no unprefixed sibling, so this is the only way to reach those models (e.g. glm-5, glm-4p7, qwen3-vl-30b-a3b-instruct, deepseek-v3p1).
- Remaining 14 entries have unprefixed siblings and become truly redundant; can be marked deprecated in a follow-up.

Verified locally: calls to
  fireworks_ai/accounts/fireworks/models/glm-5 return 404 before the patch and 200 after, hitting the same upstream as accounts/fireworks/models/glm-5.

Made-with: Cursor

### DIFF
--- a/packages/proxy/src/proxy.ts
+++ b/packages/proxy/src/proxy.ts
@@ -2233,6 +2233,13 @@ async function fetchOpenAI(
     delete bodyData["stream_options"];
   }
 
+  if (secret.type === "fireworks" && typeof bodyData.model === "string") {
+    // Strip LiteLLM-style "fireworks_ai/" prefix before forwarding; Fireworks'
+    // native API only recognizes ids like "accounts/fireworks/models/glm-5".
+    // Catalog entries sourced from LiteLLM keep the prefix (see sync_models.ts).
+    bodyData.model = bodyData.model.replace(/^fireworks_ai\//, "");
+  }
+
   if (
     secret.type === "mistral" ||
     secret.type === "databricks" ||


packages/proxy/src/proxy.ts
```diff
@@ -2233,6 +2233,13 @@ async function fetchOpenAI(
     delete bodyData["stream_options"];
   }
 
+  if (secret.type === "fireworks" && typeof bodyData.model === "string") {
+    // Strip LiteLLM-style "fireworks_ai/" prefix before forwarding; Fireworks'
+    // native API only recognizes ids like "accounts/fireworks/models/glm-5".
+    // Catalog entries sourced from LiteLLM keep the prefix (see sync_models.ts).
+    bodyData.model = bodyData.model.replace(/^fireworks_ai\//, "");
+  }
+
```
ankrgyl, line 2242: this logic should be done in the sync script, not in the proxy itself

PR #457, opened 2026-04-21
Auto update models

ankrgyl:


cpinn (approved):
Thank you for fixing.

.github/workflows/sync-models.yaml
```diff
@@ -53,8 +59,20 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Update model list
-        run: pnpm dlx tsx packages/proxy/scripts/sync_models.ts update-models --write
```
cpinn, line 57: did we also want to keep the update models in? I have seen a few times where pricing might get updated and we miss it.

.github/workflows/sync-models.yaml
```diff
@@ -1,12 +1,17 @@
...
 on:
   schedule:
-    - cron: "0 6 * * *"
+    - cron: "0 * * * *"
   workflow_dispatch:
+    inputs:
+      providers:
+        description: "Space-separated provider allowlist"
+        required: false
+        default: "fireworks openai anthropic baseten"
```
cpinn: should we add grok? or do we have some issue with those models.

packages/proxy/schema/model_list.json
```diff
@@ -5893,7 +6042,7 @@
       "xAI"
     ]
   },
-  "fireworks_ai/accounts/fireworks/models/deepseek-coder-v2-instruct": {
+  "accounts/fireworks/models/deepseek-coder-v2-instruct": {
```
aswink: I think we have to drop the "accounts/" and the "models/", this should be something like `fireworks/deepseek-coder-v2-instruct` https://fireworks.ai/models/fireworks/deepseek-coder-v2-instruct

.github/workflows/sync-models.yaml
ankrgyl: happy to

.github/workflows/sync-models.yaml
ankrgyl: yes

packages/proxy/schema/model_list.json
ankrgyl: their docs are wrong, actually. if you look at the code snippets, they are all prefixed with `accounts/`.

.github/workflows/sync-models.yaml
ankrgyl: sure yeah

PR #459, opened 2026-04-21
fix cache tokens, embeddings and resolved model

cpinn:


packages/proxy/src/proxy.ts
```diff
@@ -826,23 +826,31 @@ export async function proxyV1({
                   result.usage,
                 );
                 if (extendedUsage.success) {
+                  const legacyPromptCacheMetrics = getLegacyPromptCacheMetrics(
```
ibolmo: ah yes.. this will be interesting. i was hoping at this point we'd have normalized metrics. did you find that certain models/providers the metrics were different (legacy?)?

PR #474, opened 2026-04-23
Properly nest and name models

cpinn:
A number of our models should have a parent to nicely nest versions and have a pretty display name


aswink (approved):
🙇

PR #499, opened 2026-04-28
improve the sync_model and model finder jobs

cpinn:
- modify the existing prompts to be more targeted both when finding and fixing particular issues
- add an llm step to the sync_models job to add a display name
- remove all embedding models. Embedding models do not work in the playground which will auto route to chat completions. Ensure no embedding models are pulling in through sync_model or model finder.

With embedding models I get the following error which makes sense since they are currently being routed to the chat/completions endpoint.
`"message": "You are not allowed to sample from this model",`

You can still call embeddings directly but they do not need to be part of the model registry.


packages/proxy/schema/index.ts
```diff
@@ -818,10 +814,6 @@ export const AvailableEndpointTypes: { [name: string]: ModelEndpointType[] } = {
   "sora-2": ["openai", "azure"],
   "sora-2-pro": ["openai", "azure"],
   "sora-2-pro-high-res": ["openai", "azure"],
-  "text-embedding-3-large": ["openai", "azure"],
```
cpinn, line 821: These were added during the large sync-everything change. 

We only allow through chat and completions from the playground. We don't really have a view for embeddings at the moment, they need to be called directly. 

Removing them from the list.

aswink (approved):
nice!

PR #558, opened 2026-05-13
allow passing fetch to specific providers

ibolmo:


packages/proxy/src/proxy.ts
```diff
@@ -2689,6 +2694,7 @@ async function fetchAnthropicMessages({
         modelSpec,
         body,
         signal,
+        fetch: customFetch,
```
ibolmo, line 2747: makes sure if upstream we need to override the fetch that we're passing that custom fetch everywhere

erin2722, line 2747: I think we are missing the custom fetch in one more callsite here: https://github.com/braintrustdata/braintrust-proxy/blob/7f560a67d290bd7f0cb5b97caac6c6aa01beef7b/packages/proxy/src/proxy.ts#L3180

ibolmo: nice ty

erin2722 (approved):
Sweet, thanks!!!

erin2722:
the js/tests are flaky and we need to fix them, feel free to merge

PR #723, opened 2026-06-02
Batch model sync and bot fixes daily, and add a round of codex review

erin2722:
Having a separate PR for every model update and fix was becoming unwieldy, primarily because each update touches `index.ts` and causes merge conflicts.

This PR sets the schedule for both of these bots to daily, and makes the PRs appear in a daily batch rather than as individual changes.

I separately have a starfolk box that is configured to do manual model verification, which I plan to eventually automate to run on each of these PRs and post in a slack channel.


cpinn:
sounds great, you can probably delete the other verification jobs with the starfolk replacement!

cpinn (approved):

PR #728, opened 2026-06-03
Support openai wif

knjiang:


knjiang:
* **#728** <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/braintrust-proxy/728?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/proxy/schema/index.test.ts
```diff
@@ -344,6 +344,39 @@ describe("APISecretSchema compatibility", () => {
...
+      future_field: "future-value",
+    });
+  });
+
+  it("rejects raw OpenAI workload identity metadata", () => {
+    const result = APISecretSchema.safeParse({
+      secret: "__OPENAI_WIF__",
+      type: "openai",
+      metadata: {
+        auth_type: "workload_identity_federation",
```
erin2722: is this what we want? the other providers do accept WIF as an auth_type -- its not gonna be the biggest deal either way right now, because on the control plane we transform the auth_type into bearer before storing, but if this accidentally got saved to the data plane, it would break all secret parsing.

mostly though, we probably want to keep behavior consistent across the different providers.

knjiang: oh yeah, forgot about that. let me change it back so we are consistnet

erin2722 (approved):

PR #768, opened 2026-06-09
fix(ci): run model-sync scripts with pinned local tsx, not pnpm dlx

erin2722:
The "sync new models" job failed with `Cannot find module 'zod'` when running sync_models.ts. The scripts were invoked via `pnpm dlx tsx`, which (1) fetches an unpinned tsx into an isolated dlx store on every run, and (2) resolves the script's bare imports outside the workspace. `zod` is a dependency of `packages/proxy` and is not hoisted to the repo root, so a tsx build whose `resolveTsPaths` no longer falls back to the package's node_modules (the proxy tsconfig sets `paths` without a `baseUrl`) cannot resolve it.

Pin `tsx` as a root devDependency and invoke the scripts via `pnpm exec tsx` — the same in-workspace pattern already used for `pnpm exec vitest` in these workflows. This makes the tsx version deterministic and resolves `zod` from the installed workspace tree. Applied across sync-models, fix-missing-model-bot-issues, and verify-deployed-models (all shared the same dlx pattern).

Verified locally on Node 24.16.0 (the CI version): `pnpm exec tsx packages/proxy/scripts/sync_models.ts normalize-local-models` runs to completion (rc=0), loading the full zod-based schema.

### DIFF
--- a/.github/workflows/fix-missing-model-bot-issues.yaml
+++ b/.github/workflows/fix-missing-model-bot-issues.yaml
@@ -236,7 +236,7 @@ jobs:
             body_file="$BATCH_DIR/issue-${issue_number}.body.md"
             result_file="$BATCH_DIR/results/issue-${issue_number}.initial.json"
 
-            pnpm dlx tsx packages/proxy/scripts/fix_bot_issue.ts resolve-issue \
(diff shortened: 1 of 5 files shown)


erin2722:
fixing https://github.com/braintrustdata/braintrust-proxy/actions/runs/27204770376/job/80382448871#step:22:19

PR #782, opened 2026-06-12
Introduce concept of equivalent model names

erin2722:
see https://github.com/braintrustdata/braintrust/pull/15721


packages/proxy/schema/model_list.json
erin2722, line 1: this was all auto-modified via sync-models

knjiang (approved):

PR #818, opened 2026-06-17
update claude naming order

annalinwang:
claude display names should have the number at the end (e.g. Claude Opus 4.8 instead of Claude 4.8 Opus)
CleanShot 2026-06-17 at 17 54 08@2x


erin2722 (approved):
ty 💚

PR #864, opened 2026-06-25
Add script to deprecate models

erin2722:
LiteLLM seems to be falling behind on deprecating models, and a few customers have mentioned this as a code smell for our product. This adds a new script that calls the model list endpoints on the providers that have it available (eventually I might just start treating this as the canonical source all around, but this will be a nice way to test it out). It also changes the way we make provider calls in these scripts to centralize on using the Braintrust CI key, because that way we only need to manage a single set of keys for all the model provider stuff.


packages/proxy/scripts/deprecated_model_ids.json
```diff
@@ -0,0 +1,82 @@
+[
```
erin2722, line 1: this protects against the sync models or ai-powered update model jobs from re-adding these old models if other online sources are out of date

knjiang (approved):

knjiang:
https://platform.claude.com/docs/en/about-claude/model-deprecations

PR #1267, opened 2026-09-17
feat: Add typesafe.ai as provider and automation

lforst:
https://docs.typesafe.ai/models


erin2722 (commented):
thanks! we will have to add gateway support for this before we can advertise the models as available in our model list, @CLowbrow is adding a temporary/hacky solution to allow typesafe for scorers today and then after that I think we can merge this in tandem with gateway PRs that add support.

lforst:
@erin2722 my main objective here was having price data on typesafe.ai spans created through the SDK. Do we really need to add gateway support before merging this? I wasn't expecting the systems to be so interconnected while unrelated.

erin2722:
ahhhh I see now, I actually didn't know that the SDK pulls prices from here as well, but I suppose that makes sense. yeah there is some really annoying intermingling here, let me just get a main repo PR up that will make sure we don't accidentally expose typesafe as a provider before we support it

lforst:
@erin2722 Amazing ty!

The SDK doesn't pull prices directly but the backend does based on the provider and model the SDK sends in the metadata.

erin2722 (approved):

erin2722:
ok we can merge this and then will just update the main repo with https://github.com/braintrustdata/braintrust/pull/20608 so that this doesn't show up in our model pickers and whatnot before it is supported by @CLowbrow 's work

